PR #3, opened 2026-05-05
Add Cloudflare and GoDaddy DNS examples

buddy-n8n:
Adds two new complete deployment examples — Cloudflare DNS and GoDaddy DNS — alongside an architectural cleanup that keeps the root module AWS-only.
                                                                                                                        
## What's new                                                                                                                                                                                                                                                                                                                                                             
**examples/cloudflare/** — end-to-end example using Cloudflare for DNS validation and the ALB CNAME record. Manages its own ACM certificate lifecycle and passes the validated `certificate_arn` into the module. 
**examples/godaddy/** — same structure using the `veksh/godaddy-dns` provider. Credentials are wired through plain sensitive variables (`godaddy_api_key` / `godaddy_api_secret`) because provider blocks cannot reference dynamic data sources such as AWS Secrets Manager.
                                                                                                                                               
## Root module changes
Previously the module owned ACM + DNS resources for all three providers (Route53, Cloudflare, GoDaddy). This forced every caller to have all three providers configured, which broke `terraform apply` for single-provider deployments Cloudflare's provider requires explicit credentials at init time even when its resources are count = 0.                                                                                                                                            
                                                                                                                                                                                        
The module is now AWS-only:                                                                                                                                                            
- Removed `cloudflare_zone_id` variable and all Cloudflare resources
- Removed cloudflare and godaddy-dns from `required_providers`
- Simplified the DNS input to `certificate_arn` vs `route53_zone_id`
- Cloudflare and GoDaddy examples each own their ACM cert and DNS records, then pass `certificate_arn` to the module 

## Testing                                                                                                                                                                                                                    
All four targets (root, examples/complete, examples/cloudflare, examples/godaddy) pass `terraform validate`, `terraform test`, `tflint`, and `terraform-docs --output-check`. The cloudflare and godaddy example tests carry a NOTE explaining why full plan-level tests aren't feasible: `for_each` over `domain_validation_options` (computed after apply) cannot be exercised under a mocked AWS provider. Coverage for module wiring is provided by the root module test suite via the `certificate_arn` path.


jrx:
hey @buddy-n8n, I added a few tests and double-checked that the examples/complete still runs end to end, full cycle and with smoke-test. 

Would mind you doing this the same with the Cloudflare and GoDaddy examples to just confirm the deploys still go through, after my changes? if that looks good, I’m ready to merge 😁

buddy-n8n:
@jrx  I was able to deploy successfully with the new changes 👍

jrx (approved):

PR #4, opened 2026-05-08
Examples/medium and large

buddy-n8n:
examples/medium (~5-15M executions/day, ~60-175 req/s):                                                                                                                                                                                                                                                                                                                                                                        
  - 5× m6i.2xlarge nodes (8 vCPU, 32 GB), /24 private subnets, single NAT                                                                                                                                                                                                                                                                                                                                                        
  - RDS db.m6g.2xlarge Multi-AZ, 200 GB gp3                                                                                                                                                                                                                                                                                                                                                                                      
  - Webhook HPA min=5/max=50, worker KEDA min=5/max=40, concurrency=20                                                                                                                                                                                                                                                                                                                                                           
  - 7-day pruning, 2 Gi webhook memory limit, ElastiCache cache.r6g.large                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                  
examples/large (~50-60+M executions/day, ~350-960 req/s):                                                                                                                                                                                                                                                                                                                                                                        
  - 10× m7i.4xlarge nodes (16 vCPU, 64 GB), /20 private subnets (4,094 IPs                                                                                                                                                                                                                                                                                                                                                       
    each — /24s exhausted by VPC CNI warm pools at 20+ large nodes), dual NAT                                                                                                                                                                                                                                                                                                                                                    
  - Aurora PostgreSQL 16.4 I/O-Optimized (writer + reader, db.r6g.8xlarge),                                                                                                                                                                                                                                                                                                                                                      
    removes the IOPS ceiling that caps RDS gp3 around 600 req/s                                                                                                                                                                                                                                                                                                                                                                  
  - VPC CNI addon with WARM_ENI_TARGET=0 / WARM_IP_TARGET=2                                                                                                                                                                                                                                                                                                                                                                      
  - PgBouncer (2 replicas, transaction mode) deployed at the example level —                                                                                                                                                                                                                                                                                                                                                     
    without pooling, 80 webhook + 160 worker + 2 main = 1,210 connections                                                                                                                                                                                                                                                                                                                                                        
    approaches Aurora's max                                                                                                                                                                                                                                                                                                                                                                                                      
  - Webhook HPA min=30/max=80, worker KEDA min=20/max=160, concurrency=40                                                                                                                                                                                                                                                                                                                                                        
  - 24h pruning, 4 Gi webhook memory, n8n helm timeout 1800s                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                  
README.md: new Examples section comparing complete/medium/large across                                                                                                                                                                                                                                                                                                                                                           
22 dimensions (target scale, throughput, compute, DB class + IOPS,                                                                                                                                                                                                                                                                                                                                                               
pooling, autoscaling bounds, retention, cost). Section sits outside the                                                                                                                                                                                                                                                                                                                                                          
terraform-docs inject markers so it survives README regeneration.                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                  
Root module changes (so the same module drives both managed RDS and                                                                                                                                                                                                                                                                                                                                                              
external-Aurora deployments):                                                                                                                                                                                                                                                                                                                                                                                                    
  - New variables: create_database, db_host, db_password,                                                                                                                                                                                                                                                                                                                                                                      
    db_postgresdb_pool_size, db_postgresdb_ssl_enabled, n8n_helm_timeout                                                                                                                                                                                                                                                                                                                                                         
  - aws_db_instance.n8n + aws_db_subnet_group.n8n now gated on                                                                                                                                                                                                                                                                                                                                                                   
    create_database (a static bool — db_host is computed-known-after-apply                                                                                                                                                                                                                                                                                                                                                       
    and cannot drive a count expression)                                                                                                                                                                                                                                                                                                                                                                                         
  - n8n.tf and outputs.tf use create_database for short-circuit-safe                                                                                                                                                                                                                                                                                                                                                             
    conditional references to the optional RDS resources                                                                                                                                                                                                                                                                                                                                                                         
  - n8n helm release: DB host conditional on create_database, SSL env vars                                                                                                                                                                                                                                                                                                                                                       
    conditional on db_postgresdb_ssl_enabled (false when the DB host is an                                                                                                                                                                                                                                                                                                                                                       
    in-cluster pooler that handles SSL upstream),                                                                                                                                                                                                                                                                                                                                                                                
    QUEUE_BULL_REDIS_KEEP_ALIVE=true added to prevent ElastiCache                                                                                                                                                                                                                                                                                                                                                                
    idle-disconnect crashes under sustained load, DB_POSTGRESDB_POOL_SIZE                                                                                                                                                                                                                                                                                                                                                        
    plumbed through                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                  
PgBouncer is intentionally NOT part of the root module as it's a large-scale                                                                                                                                                                                                                                                                                                                                                      
concern and lives in examples/large/pgbouncer.tf. The root module's                                                                                                                                                                                                                                                                                                                                                              
integration points are db_host + db_postgresdb_ssl_enabled. Uses                                                                                                                                                                                                                                                                                                                                                                 
edoburu/pgbouncer:v1.23.1-p3; the official pgbouncer/pgbouncer Docker Hub                                                                                                                                                                                                                                                                                                                                                        
image was last released 2020-11-22 against PgBouncer 1.15.0 and is no                                                                                                                                                                                                                                                                                                                                                            
longer maintained. AUTH_TYPE=plain stores the password as plaintext in                                                                                                                                                                                                                                                                                                                                                           
userlist.txt so PgBouncer can negotiate scram-sha-256 upstream to Aurora                                                                                                                                                                                                                                                                                                                                                         
16; AUTH_TYPE=md5 fails with "wrong password type" because the hashed-only                                                                                                                                                                                                                                                                                                                                                       
userlist cannot satisfy a SCRAM challenge. PgBouncer runs in its own                                                                                                                                                                                                                                                                                                                                                             
`pgbouncer` namespace so the example doesn't have to wait on any                                                                                                                                                                                                                                                                                                                                                                 
module-internal resource (which would create a dependency cycle through                                                                                                                                                                                                                                                                                                                                                          
the module's helm_release.n8n).                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                  
Tests:                                                                                                                                                                                                                                                                                                                                                                                                                           
  - tests/defaults.tftest.hcl: count-based RDS resource references updated                                                                                                                                                                                                                                                                                                                                                       
    to [0]; new external_db_skips_rds_instance run validates create_database                                                                                                                                                                                                                                                                                                                                                     
    = false produces no RDS resources                                                                                                                                                                                                                                                                                                                                                                                            
  - examples/medium and examples/large: cluster_name validation tests                                                                                                                                                                                                                                                                                                                                                            
  - examples/cloudflare and examples/godaddy: comment in test file rewritten                                                                                                                                                                                                                                                                                                                                                     
    to correctly explain why no run blocks are present (variable validation                                                                                                                                                                                                                                                                                                                                                      
    fires alongside the unavoidable for_each error from the example-level                                                                                                                                                                                                                                                                                                                                                        
    cert validation records, so neither approach yields a useful test)                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                  
All sizing decisions are derived from internal load testing. The large                                                                                                                                                                                                                                                                                                                                                           
example was apply/destroy-tested end-to-end against AWS during development.


jrx:
@buddy-n8n Just FYI 
I've noticed this PR #4 also has the changes from #3 so I'll merge #3 first than will do rebase and review of #4 after that.

jrx:
# Deferred follow-ups from this PR

During review + live validation (`apply` + `smoke-test` + `destroy` against all three sizing tiers on real AWS) the following items were identified but kept out of scope so this PR stays focused. Filing here as a structured comment because Issues are disabled on this repo — each section below is sized for a small follow-up PR.

---

## 1. Add cross-variable validation on `db_host` / `db_password` when `create_database = false`

**Why.** This PR adds `var.create_database` (default `true`) plus optional `var.db_host` / `var.db_password` for the external-DB path used by `examples/large/` (Aurora + PgBouncer). When a caller sets `create_database = false` but forgets to also set `db_host` and/or `db_password`, `terraform plan` succeeds and the failure surfaces deep inside the n8n Helm release at apply time — confusing to diagnose, and well after expensive resources (EKS, RDS-or-Aurora) have already been created.

**Fix.** `versions.tf` already pins `terraform >= 1.9`, so cross-variable `validation` blocks are available. Add to `variables.tf`:

```hcl
variable "db_host" {
  # …existing fields…

  validation {
    condition     = var.create_database || var.db_host != null
    error_message = "db_host is required when create_database = false."
  }
}

variable "db_password" {
  # …existing fields…

  validation {
    condition     = var.create_database || var.db_password != null
    error_message = "db_password is required when create_database = false."
  }
}
```

**Acceptance.**
- [ ] Validations added to `db_host` and `db_password` in `variables.tf`
- [ ] New plan-time test in `tests/defaults.tftest.hcl` asserts both validations fire (`expect_failures = [var.db_host]` / `[var.db_password]` with `create_database = false` and the relevant input unset)
- [ ] `terraform-docs` regenerated

---

## 2. Harden Aurora cluster defaults in `examples/large/`

**Why.** `aws_rds_cluster.n8n` in `examples/large/aurora.tf` is missing the same hardening defaults Checkov soft-fails on for the module's existing RDS instance. The module's RDS gaps are a pre-existing call (`soft_fail: true` is set in CI with a comment), but adding a *new* RDS-family resource is the right time to set the right defaults — `AGENTS.md` explicitly says: *"When you add new resources, do not regress curated findings; prefer fixing them over adding suppressions."*

**Fix.** In `examples/large/aurora.tf`, add to `aws_rds_cluster.n8n`:

```hcl
storage_encrypted                   = true
iam_database_authentication_enabled = true
enabled_cloudwatch_logs_exports     = ["postgresql"]
```

`storage_encrypted = true` is free, defaults to false on Aurora unless explicitly set, and is Registry-table-stakes for any new RDS-family resource.

**Acceptance.**
- [ ] Three attributes set on `aws_rds_cluster.n8n`
- [ ] `checkov` no longer reports `CKV_AWS_133` / `CKV_AWS_327` against the new Aurora cluster (verify in CI output)
- [ ] No regression in `terraform test` for `examples/large/`

---

## 3. Thicken plan-time asserts for `examples/large/`

**Why.** `examples/large/tests/defaults.tftest.hcl` today only asserts `cluster_name` length validation and `aurora_instance_class` regex. Several architecture facts that *should* be locked down at plan time — PgBouncer required anti-affinity (the live-validated fix in the third commit on this PR), PDB existence, Aurora writer+reader `cluster_identifier` wiring, `aws_eks_addon.vpc_cni.configuration_values` (`WARM_IP_TARGET=2`), `module.vpc.natgw_ids` length (dual-NAT-GW HA) — currently rely on live-apply validation only.

**Blocker.** The module's `dns.tf` does `for_each` over `aws_acm_certificate.n8n[0].domain_validation_options`, which is unknown at plan time under a mocked AWS provider. `mock_resource "aws_acm_certificate"` with `defaults` and top-level `override_resource { target = module.n8n.aws_acm_certificate.n8n[0] }` both silently no-op for this specific computed-set-of-object attribute (verified locally with Terraform 1.15.x). Same root cause documented in `examples/cloudflare/tests/defaults.tftest.hcl`.

**Two paths to unblock, pick one.**

**(a) Expose an optional `certificate_arn` variable on the example** (default `null`; passed through to the module). This is a genuine BYO-cert feature (useful for callers with wildcard ACM certs across multiple deployments), and the test can set it to a stub ARN to bypass the dns.tf automated path entirely. ~10 lines of example code, unlocks all asserts.

**(b) Wait for upstream Terraform mock framework to support overriding computed attributes used in `for_each`.** Track in <https://github.com/hashicorp/terraform/issues>. Lower-effort but indefinite.

**Acceptance (option a).**
- [ ] `certificate_arn` added to `examples/large/variables.tf` (default `null`)
- [ ] `examples/large/main.tf` passes it through to the module
- [ ] Test sets `certificate_arn = "arn:aws:acm:us-east-1:123456789012:certificate/test"` to bypass dns.tf
- [ ] New `run` blocks assert: dual-NAT-GW, Aurora topology (engine, storage_type, writer+reader wiring, instance_class propagation), PgBouncer namespace/replicas/required anti-affinity/PDB/ClusterIP service, `vpc-cni` addon `WARM_*` config
- [ ] Comment in the tftest file explaining why the test uses `certificate_arn` even though the example's production usage uses `route53_zone_id`

---

## 4. Update stale doc references to the example layout

**Why.** Two pieces of documentation pre-date the addition of the medium and large examples and are now inaccurate:

1. `README.md` — "Three runnable examples ship with the module" → there are now 5 (`complete`, `medium`, `large`, `cloudflare`, `godaddy`)
2. `AGENTS.md` — `examples/complete/` described as "the canonical full deployment showing every important feature" → false the moment `large` introduced Aurora + PgBouncer + VPC CNI tuning + dual-NAT-GW HA

**Fix.**

- `README.md`: rephrase to "three sizing tiers (`complete`, `medium`, `large`) plus two DNS-variant examples (`cloudflare`, `godaddy`) showing the same `complete` sizing with a different DNS provider." Consider a top-level matrix table mapping (size × DNS provider) → example folder.
- `AGENTS.md` line about `examples/complete/`: change to *"`examples/complete/` is the smallest viable production deployment using module defaults; `medium` and `large` show progressively scaled-up reference architectures; the canonical reference is whichever matches your scale."*

Related but out-of-scope for this issue (worth a separate discussion): renaming `examples/complete/` → `examples/small/` to match the size trilogy. Tradeoffs already discussed in PR #4 review.

**Acceptance.**
- [ ] README section "Three runnable examples ship with the module" updated
- [ ] AGENTS.md line about canonical full deployment updated
- [ ] `terraform-docs` output check still clean

---

## 5. Document macOS `mDNSResponder` NXDOMAIN caching after example apply→destroy→apply

**Why.** Reproducibly hit during PR #4 validation when the same FQDN (`n8n.example.com`) is reused across consecutive apply→destroy→apply cycles on the same workstation. macOS's `mDNSResponder` caches the NXDOMAIN response from the destroy phase and refuses to refresh for 5–15 minutes after the alias record reappears. `dig` and `host` (which bypass `mDNSResponder`) resolve fine; `curl`, browsers, and anything else using `getaddrinfo()` silently fail with `CURLE_COULDNT_RESOLVE_HOST` (HTTP 000). `tests/scripts/smoke-test.sh` reports these as "HTTP 000 — connection failed", which is technically correct but misleads users into thinking the deployment is unhealthy when it's actually a workstation DNS cache issue.

**Fix.** Add ~10 lines to `docs/troubleshooting.md`:

````markdown
### Smoke test reports "HTTP 000" after a recent destroy + re-apply

**Symptom.** The smoke test fails the HTTP health, redirect, and API checks
with `HTTP 000`. Direct `dig jrx-test.example.com` resolves correctly, but
`curl https://jrx-test.example.com/healthz` exits with code 6 (`CURLE_COULDNT
_RESOLVE_HOST`).

**Cause (macOS).** `mDNSResponder` cached the NXDOMAIN response from the
previous deployment's destroy phase and is serving it for 5–15 minutes even
after Terraform re-created the Route 53 alias record. `dig` and `host` bypass
`mDNSResponder`; `curl`, browsers, and anything using `getaddrinfo()` don't.

**Fix.**

```bash
sudo killall -HUP mDNSResponder
```

Or wait for the negative cache to age out (typically 5–15 min). To avoid
hitting this entirely, use a fresh subdomain per deployment.
````

Optionally, sharpen the error message in `tests/scripts/smoke-test.sh` so the "HTTP 000" path mentions this case on macOS.

**Acceptance.**
- [ ] Section added to `docs/troubleshooting.md`
- [ ] (Optional) `smoke-test.sh` error message references `mDNSResponder` on macOS

---

## 6. Tighten `AGENTS.md` local development loop to include `tflint` per example

**Why.** PR #4 originally landed with a `tflint` failure on `examples/medium` and `examples/large` (missing `versions.tf` → `terraform_required_version` and `terraform_required_providers` warnings). The local CI loop documented in `AGENTS.md` was followed faithfully but didn't catch it: the "Repeat under each example" instruction lists `init` / `validate` / `test` only, while CI (correctly) runs `tflint` per example matrix target too. Net result: a CI-only failure mode for any future contributor doing exactly what `AGENTS.md` tells them to do.

**Fix.** In the *Local development loop* section (around line 195 of `AGENTS.md`), add `tflint` to each per-example invocation so the local loop is a faithful subset of CI:

```diff
 # Repeat under each example.
-cd examples/complete && terraform init -backend=false && terraform validate && terraform test -verbose
-cd examples/cloudflare && terraform init -backend=false && terraform validate && terraform test -verbose
-cd examples/godaddy && terraform init -backend=false && terraform validate && terraform test -verbose
+cd examples/complete   && terraform init -backend=false && terraform validate && terraform test -verbose && tflint --init && tflint --format compact
+cd examples/cloudflare && terraform init -backend=false && terraform validate && terraform test -verbose && tflint --init && tflint --format compact
+cd examples/godaddy    && terraform init -backend=false && terraform validate && terraform test -verbose && tflint --init && tflint --format compact
+cd examples/medium     && terraform init -backend=false && terraform validate && terraform test -verbose && tflint --init && tflint --format compact
+cd examples/large      && terraform init -backend=false && terraform validate && terraform test -verbose && tflint --init && tflint --format compact
```

Lines for `medium` and `large` belong in this section too — they got covered by the CI matrix in PR #4 (commit `2fb8f49`) but the doc never caught up.

While editing this section, also consider adding the `terraform-docs --output-check` call per example, mirroring the CI `docs` job. This was likewise CI-covered for `medium`/`large` but not in `AGENTS.md`'s local loop. Without it, contributors only run `terraform-docs` at the module root — example READMEs drift silently until CI yells.

**Acceptance.**
- [ ] `AGENTS.md` local development loop block updated for all 5 examples
- [ ] `terraform-docs --output-check` line added (optional but recommended — same drift risk)
- [ ] The "Quality bar" section's claim that *"the suite runs without AWS credentials and is safe to run in CI"* still holds — none of the additions need AWS creds (tflint is static; terraform-docs reads source only)

This is naturally a candidate to land alongside follow-up #4 (the other `AGENTS.md` / README accuracy items) in a single docs-cleanup PR.

---

## Out of scope / parking lot

- **Rename `examples/complete/` → `examples/small/`.** Discussed in PR #4 review. Naming clarity win, ~30 line mechanical sweep, but the call has external-facing implications (the `complete` path may already be referenced in blog posts or internal docs). Best left to a focused rename PR.
- **PgBouncer image digest pinning** (`edoburu/pgbouncer:v1.23.1-p3` → digest reference). Reproducibility hygiene, low priority — the n8n chart itself uses tags throughout.
- **Aurora password URL-reserved character review** in `aurora.tf`'s `random_password` override set (`!#$%&*()-_=+[]{}<>:?` includes `:` and `?`). Today the password only flows through env vars and PgBouncer's `userlist.txt`, so no URL parsing path. Documented gotcha for any future maintainer who decides to plumb the password into a `postgres://` URI.

jrx (approved):

PR #65, opened 2026-08-04
feat: restrict ALB ingress via alb_inbound_cidrs and prefix lists

### LINKED ISSUE #42: feat: restrict ALB ingress via alb_inbound_cidrs / prefix lists
## Problem

The n8n Ingress is created `internet-facing` (`n8n.tf`, `alb.ingress.kubernetes.io/scheme = internet-facing`) with no source restriction, so the ALB accepts connections from any IP on the internet. The as-built security posture documents "only the ALB is internet-facing" but does not restrict *who* can reach it — there is no ADR covering ALB ingress source filtering, so this is an uncovered gap rather than a deliberate decision.

Many production deployments need the editor UI / REST API reachable only from a known set of networks (corporate egress ranges, VPN, office CIDRs, partner prefixes) while still terminating on a public ALB. Today that requires forking the module or attaching rules out of band.

## Proposal

Add opt-in variables to restrict ALB ingress at the source:

- `alb_inbound_cidrs` (list(string), default `[]` = unrestricted, current behavior) rendered into `alb.ingress.kubernetes.io/inbound-cidrs`.
- Optionally `alb_inbound_prefix_list_ids` (list(string), default `[]`) for managed prefix lists, rendered into the corresponding LBC annotation / security-group rule.

When both are empty the annotation is omitted entirely and the ALB stays open (no plan diff for existing callers). Validation should reject malformed CIDRs.

## Notes

- Purely additive and backward compatible; default preserves current open behavior.
- Follows the curated-surface contract (dedicated, validated, documented variables).
- Consider documenting the interaction with `internet-facing` scheme (this narrows a public ALB; it is not the same as switching to an internal ALB).
- Needs a `variables.tf` entry with validation, wiring in `n8n.tf` (Ingress annotations), a `.tftest.hcl` assert, a CHANGELOG bullet, and a `terraform-docs` refresh.

buddy-n8n:
## Summary
- Add `alb_inbound_cidrs` and `alb_inbound_prefix_list_ids` inputs, rendered into the AWS Load Balancer Controller's `inbound-cidrs` and `security-group-prefix-lists` annotations. Both default to an empty list and omit the annotation entirely when unset, so no existing deployment sees a plan diff.
- IPv4-only and no-host-bits validation on `alb_inbound_cidrs`, matching what the controller and EC2 actually accept.
- Three plan-time warnings cover ways a restriction can silently not apply: `ingress_annotations` overriding the same key, a caller-owned `security-groups` annotation, and either input set with `create_ingress = false`.
- `docs/troubleshooting.md` documents a fourth path (IngressClassParams) that cannot reach the module-managed Ingress but can reach caller-owned ones like `examples/split-ingress`.
- `examples/split-ingress`'s `admin_allowed_cidr_blocks` now gets the same no-host-bits guard it was missing.
- This restriction covers the whole ALB, webhook paths included, since the module builds a single ALB for both editor and webhooks. Docs and variable descriptions point at `examples/split-ingress` for splitting the two.

## Test plan
- [x] Live-tested against a real EKS cluster running LBC v3.5.0: confirmed `alb_inbound_cidrs` and `alb_inbound_prefix_list_ids` apply correctly, restriction covers port 80 and 443, CIDRs/prefix lists union, lockout presents as a client timeout, and recovery is a `terraform apply` (LBC reverts manual security-group edits).
- [x] Two claims from an earlier draft of this change were disproved by that live test and corrected before this PR: an empty annotation value does not black-hole the ALB (LBC treats an empty parsed list the same as an absent annotation), and IngressClassParams cannot override the module-managed Ingress (it's classified via the legacy `kubernetes.io/ingress.class` annotation, which the controller matches before ever loading IngressClassParams).
- [x] `terraform fmt -check -recursive`, `terraform init`, and `terraform validate` pass on the root module and all six examples (`small`, `medium`, `large`, `cloudflare`, `godaddy`, `split-ingress`).
- [x] `terraform test -verbose` passes on the root module and all six examples.
- [x] `tflint --format compact` passes on the root module and all six examples (0 findings).
- [x] `terraform-docs --output-check` passes on the root module and all six examples (checked against CI's pinned v0.22.0).
- [x] `checkov` findings unchanged vs `main` (23 pre-existing, unrelated findings; `soft_fail: true` regardless).

Closes #42


jrx (dismissed):
Reviewed against the Terraform Registry / Partner Premier quality bar and the repo's AGENTS.md, with all review findings resolved on the branch:

- Rebased cleanly onto current main (linear history, no merge commits); the review fix (nullable = false on both new inputs) was squashed into the feature commit with original authorship preserved.
- Full local loop green at root and all six examples: fmt, validate, terraform test (156 runs at root), tflint, terraform-docs --output-check. CI matrix is green on the final push.
- Verified live on the pr65 cluster (examples/small, LBC v3.5.0): defaults produce no plan diff on an existing deployment; alb_inbound_cidrs narrows the controller-managed security group to exactly the given ranges on both listen ports; an allowed source gets 200/301 and an excluded source gets the documented connection timeout; alb_inbound_prefix_list_ids applies and unions with the CIDRs (reachable via a /32 present only in the prefix list); the security-groups conflict check warns on a real plan; reverting to defaults restores 0.0.0.0/0 with no residue.
- Live verification surfaced one real hazard: prefix lists count against the security-group rule quota by max-entries weight per listen port, so a heavy list (e.g. the CloudFront origin-facing list) black-holes the ALB after a clean apply via RulesPerSecurityGroupLimitExceeded. Documented in the variable description, README caveats, troubleshooting guide, and CHANGELOG in the final commit.

Backward compatible, additive-only, secure defaults preserved. Approving.

jrx (approved):
Re-approving after 2646c61, which resolves cubic's three P3 findings: the condensed quota examples in the variable description, README caveat, and CHANGELOG now say the CloudFront origin-facing list needs 110 rules by itself, reproducible from the 2 x (weight + CIDR count) formula stated alongside; the troubleshooting entry keeps 2 x (55 + 1) = 112 because it states the single-CIDR assumption from the live incident. Docs-only change, full CI matrix green, terraform-docs in sync, root test suite passing (156).

Issue #66, opened 2026-08-04
KEDA worker autoscaler loses Redis auth when redis_transit_encryption_enabled is set


buddy-n8n:
## Problem

#41 adds an opt-in `redis_transit_encryption_enabled` variable that enables TLS + an AUTH
token on the ElastiCache Redis cluster n8n uses for its queue backend. Once that's on,
Redis requires TLS and authentication for every connection.

n8n's own pods handle both — the chart wires `redis.tls` → `QUEUE_BULL_REDIS_TLS` and
`redis.passwordSecret` → `QUEUE_BULL_REDIS_PASSWORD` onto main/worker/webhook-processor.
The KEDA worker autoscaler does not. Its trigger in `n8n.tf` has neither:

```hcl
{
  type = "redis"
  metadata = {
    address    = "<redis-host>:6379"
    listName   = "bull:jobs:wait"
    listLength = "..."
  }
  authenticationRef = { name = "" }
}
```

Against a TLS+AUTH Redis the metric read fails. Nothing crashes — the ScaledObject simply
stops tracking backlog and workers sit at a static replica count between
`n8n_worker_keda_min_replicas` and `n8n_worker_keda_max_replicas`. Silent degradation, so
the deployment still looks healthy.

Two distinct gaps, with different fixes:

| Gap | What KEDA needs |
|---|---|
| **TLS** | `enableTLS: "true"` in trigger metadata — trivial |
| **AUTH** | credentials, which metadata cannot carry as a literal |

## Proposal

The TLS half is settled. The AUTH half has two candidate approaches, and picking between
them is the first task.

### Option A — `passwordFromEnv` (verify this first)

KEDA's Redis scaler accepts `passwordFromEnv`, which names an **environment variable on the
scale target's container** rather than carrying a secret. The scale target is the n8n worker
Deployment, which after #41 already has `QUEUE_BULL_REDIS_PASSWORD` via `secretKeyRef`. If
KEDA resolves that, the whole fix is two metadata keys — no new resource:

```hcl
enableTLS       = "true"
passwordFromEnv = "QUEUE_BULL_REDIS_PASSWORD"
```

The chart needs no change: `scaledobject-worker.yaml` renders arbitrary metadata keys via
`{{- range $key, $value := .metadata }}`.

**Unverified assumption, and it is the whole option:** that KEDA's env resolver follows
`secretKeyRef` indirection, not just literal `value:` entries. KEDA's docs describe `FromEnv`
[...]

PR #67, opened 2026-08-04
feat: add redis_transit_encryption_enabled for ElastiCache TLS + AUTH

### LINKED ISSUE #41: feat: add redis_transit_encryption_enabled for ElastiCache TLS + auth token
## Problem

The module connects n8n to its ElastiCache Redis queue backend in plaintext. `redis.tf` creates the cluster with no in-transit encryption and no auth token, and `n8n.tf` sets `redis.tls = false`. The security posture is deliberately network-trust based: Redis sits in private subnets behind a security group that only admits traffic from within the VPC, so isolation is by network boundary rather than by credentials.

That default is defensible inside a trusted private VPC, and it is the accepted as-built posture (see the security-posture ADR, which explicitly weighed "ElastiCache with TLS + AUTH" and chose network isolation instead). The same ADR, however, records "ElastiCache has no auth" as a residual risk worth a customer conversation: a network-boundary breach reaches Redis unauthenticated, and the queue payloads (workflow execution data) traverse the VPC in cleartext.

There is currently no way for a caller to opt into a stricter posture without forking the module.

## Proposal

Add an opt-in, default-off `redis_transit_encryption_enabled` variable (bool, default `false` to preserve the accepted as-built behavior). When `true`:

- Enable `transit_encryption_enabled` on the ElastiCache resource.
- Generate an auth token (`random_password`, ElastiCache token constraints: 16-128 chars, restricted symbol set) and set `auth_token` on the cluster.
- Store the token in the existing n8n DB/secret pattern and wire `redis.tls = true` plus the auth password through the Helm values.

Callers who want credential-based Redis security (parity with the stance other cloud lines in this space take) can flip one switch; everyone else keeps the network-trust default with no plan diff.

## Notes

- Purely additive and backward compatible; default preserves the accepted network-trust posture.
- Follows the curated-surface contract (dedicated, validated, documented variable rather than a generic escape hatch).
- Enabling transit encryption / auth on an existing cluster may force replacement — the variable description and CHANGELOG should call that out.
- Needs a `variables.tf` entry with validation, wiring in `redis.tf` + `n8n.tf`, a `.tftest.hcl` assert, a CHANGELOG bullet, and a `terraform-docs` refresh.

buddy-n8n:
Closes #41.

Adds an opt-in, default-off `redis_transit_encryption_enabled` that puts the ElastiCache queue backend behind TLS and an AUTH token. The default preserves the module's network-trust posture, so callers who leave it alone see no plan diff.

## Why the opt-in path uses a different AWS resource

`auth_token` does not exist on `aws_elasticache_cluster`. AWS exposes it only on `aws_elasticache_replication_group`, and further requires transit encryption to be enabled before AUTH can be used at all. The module therefore carries both resources, count-gated and mutually exclusive:

| `redis_transit_encryption_enabled` | Resource |
| --- | --- |
| `false` (default) | `aws_elasticache_cluster.n8n[0]` |
| `true` | `aws_elasticache_replication_group.n8n[0]` |

Migrating everyone to the replication group would have read more simply, but it replaces the cache for every existing deployment. Both are single-node with the same `redis_node_type`, so enabling the flag changes neither topology nor cost. Consumers read `local.redis_host` rather than either resource, so the Helm values, KEDA trigger addresses and the `redis_endpoint` output cannot drift apart.

A `moved` block absorbs the `count` added to `aws_elasticache_cluster.n8n`. Without it, the new address reads as a different resource and **every consumer replaces their Redis on upgrade**, whether or not they set the new variable.

## Live verification

Applied against a real cluster (`iss41dev`, us-east-1, default sizing). Full log in the deploy harness; summary:

| Scenario | Result |
| --- | --- |
| S1 — baseline on `main` | Pass. 75 resources, six pods Running |
| S2 — upgrade to this branch, flag off | Pass. `Plan: 0 to add, 0 to change, 0 to destroy` |
| S3 — enable flag | **Failed. Found a real defect** (below) |
| S3b — enable flag, after fix | Pass. TLS + AUTH working end to end |
| S4 — KEDA limitation | Confirmed, mechanism corrected |

### S2 is the backward-compatibility proof

```
# module.n8n.aws_elasticache_cluster.n8n has moved to module.n8n.aws_elasticache_cluster.n8n[0]
Plan: 0 to add, 0 to change, 0 to destroy.
```

This is the claim a plan-time test structurally cannot make: a fresh plan writes `[0]` from the start and passes whether or not the `moved` block exists. It had to be measured against real prior state.

### The defect the live run caught

```
Error: creating ElastiCache Replication Group (iss41dev-redis):
InvalidParameterValue: Cannot have a cluster and replication group with same identifier.
```

ElastiCache shares one identifier namespace across cache clusters and replication groups. The two Terraform resources have no dependency edge, so Terraform created the replacement while the old cache still existed. It had **already destroyed the cluster** by then, leaving the deployment with no queue backend and needing a second apply to recover.

The docs at that point promised "flipping the flag replaces Redis and loses queued jobs." The truth was "destroys Redis, fails the apply, leaves nothing." Materially different advice. Fixed by naming the replication group `<cluster_name>-redis-tls`, which removes the hazard in both directions, and pinned by a regression test.

Worth noting: S2 and all 104 plan-time tests passed while this was broken.

### Proof the feature actually works

Protocol-level, from inside the cluster:

| Probe | Result |
| --- | --- |
| TLS + token | `PONG` |
| TLS + token, `KEYS bull:*` | `bull:jobs:stalled-check` |
| TLS, no token | `NOAUTH Authentication required.` |
| Plaintext, no TLS | no response |

The last two are the security claim itself: unauthenticated and unencrypted access are both refused.

End to end, with `OFFLOAD_MANUAL_EXECUTIONS_TO_WORKERS=true` forcing the job across the queue:

```
POST /rest/workflows/<id>/run  → 200, executionId 1
GET  /rest/executions/1        → status=success, output contains "redis-tls-auth-ok"

worker log:
  Worker started execution 1 (job 1)
  Worker finished execution 1 (job 1)
```

### A risk that did not materialise

n8n's config recommends `QUEUE_BULL_REDIS_DNS_LOOKUP_STRATEGY=NONE` for ElastiCache with TLS, warning of hostname/IP cert mismatches. It was deliberately not set pre-emptively so this run would produce evidence rather than a guess. No cert errors appeared, consistent with that caveat applying to cluster-mode ElastiCache resolving nodes to IPs, whereas this is cluster-mode-disabled connecting by hostname. Left unset.

## Known limitation shipped with this (#66)

KEDA's worker autoscaler cannot reach an encrypted Redis. Observed:

```
ERROR scale_handler error resolving auth params
  {"name": "n8n-worker", "error": "connection to redis failed: i/o timeout"}

keda-hpa-n8n-worker   TARGETS: <unknown>/5 (avg), <unknown>/5 (avg)   1   10   2
```

An i/o timeout, not `NOAUTH` — KEDA opens a plaintext connection to a TLS-only endpoint and hangs, never reaching authentication. So **TLS is the first-order fix for #66**, not auth; #66 has been updated with this. n8n itself is unaffected — the autoscaler is the only casualty, which is what makes it easy to miss.

Documented in the variable description, README, and CHANGELOG. Sub-issue #66 tracks the fix.

## Checks

`fmt` clean, `validate` Success, `tflint` 0 findings, `terraform-docs` in sync, **105 tests pass**, `checkov` identical to `main` (171/28/7).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,51 @@ this project adheres to the stability contract in
 
 ### Added
 
+- `redis_transit_encryption_enabled` input (default `false`) puts the
+  ElastiCache queue backend behind TLS and an AUTH token. The default preserves
+  the module's network-trust posture — Redis in private subnets behind a
+  VPC-only security group — so callers who leave it alone get no plan diff at
+  all. Callers who want credential-based Redis security flip one switch instead
+  of forking the module. Resolves
+  [#41](https://github.com/n8n-io/terraform-aws-n8n/issues/41).
+
+  The opt-in path uses a **different AWS resource**, which is worth
+  understanding before enabling it. `auth_token` does not exist on
+  `aws_elasticache_cluster`; AWS exposes it only on
+  `aws_elasticache_replication_group`, and further requires transit encryption
+  to be on before AUTH can be enabled at all. So the module now carries both
+  resources, count-gated on the variable and mutually exclusive: the default
+  path keeps the `aws_elasticache_cluster` it has always used, and only the
+  opt-in path creates a replication group (single node, `num_cache_clusters =
+  1`, matching the cluster path's topology and cost). Everything downstream —
+  Helm values, KEDA triggers, the `redis_endpoint` output — reads a shared
+  `local.redis_host` rather than either resource, so the two paths cannot
+  drift.
+
+  **Switching the flag on an existing deployment replaces Redis**, since the
+  two resources are different objects. Every job queued at that moment is lost.
+  Drain workers and pick a maintenance window. Upgrading the module *without*
+  touching the variable does **not** replace anything: a `moved` block in
+  `refactoring.tf` absorbs the `count` added to `aws_elasticache_cluster.n8n`.
+  Verified on a live cluster — a deployment applied on the previous version and
+  then repointed at this one planned `No changes.`
+
+  The generated token respects ElastiCache's constraints (16-128 chars, with
+  `! & # $ ^ < > -` the only permitted non-alphanumerics) and is published two
+  ways: as a Kubernetes secret the chart mounts as `QUEUE_BULL_REDIS_PASSWORD`
+  on main, worker and webhook processor, and as a new sensitive
+  `redis_auth_token` output (`terraform output -raw redis_auth_token`).
+
+  **Known limitation:** KEDA's worker autoscaler does not yet authenticate to
+  Redis. Its trigger carries neither TLS nor credentials, so while this flag is
+  on, queue-depth autoscaling stops responding to backlog and workers hold a
+  static count between `n8n_worker_keda_min_replicas` and
+  `n8n_worker_keda_max_replicas`. Nothing crashes, which is what makes it easy
+  to miss. Tracked in
+  [#66](https://github.com/n8n-io/terraform-aws-n8n/issues/66); until that
+  lands, treat this flag as trading queue-depth autoscaling for encryption and
+  authentication.
+
 - `n8n_license_detach_floating_on_shutdown` input (default `false`) maps to
   `N8N_LICENSE_DETACH_FLOATING_ON_SHUTDOWN`, overriding n8n's own upstream
   default of `true`. In multi-main (the module default), the leader main
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,12 +87,24 @@ this project adheres to the stability contract in
   on main, worker and webhook processor, and as a new sensitive
   `redis_auth_token` output (`terraform output -raw redis_auth_token`).
 
-  **Known limitation:** KEDA's worker autoscaler does not yet authenticate to
-  Redis. Its trigger carries neither TLS nor credentials, so while this flag is
-  on, queue-depth autoscaling stops responding to backlog and workers hold a
-  static count between `n8n_worker_keda_min_replicas` and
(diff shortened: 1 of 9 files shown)


buddy-n8n:
Superseded by #76, which resolves both #41 and #66.

Worth recording why this branch is a dead end rather than something to rebase. `fix/issue-66-keda-redis-tls-auth` (#70) branched from this one at `3032293` and already contained every commit here, then replaced the `KNOWN GAP` comment block with a working KEDA implementation. So this branch's tip commits (`d9612c6`, `c277d27`) were a first attempt at the same fix that #70 superseded.

The two also could not ship separately. Merging this alone puts a broken feature in main: enabling TLS makes KEDA open a plaintext connection to a TLS-only endpoint, where it hangs on `connection to redis failed: i/o timeout`. Nothing crashes, the HPA just reports `<unknown>` and workers freeze at their current replica count.

The substance is preserved in #76, including the identifier constraint this branch found the hard way on a live cluster (`InvalidParameterValue: Cannot have a cluster and replication group with same identifier`), which is now pinned by a test.

One change reviewers of this PR should know about: the replication group identifier is **not** `<cluster_name>-redis-tls`. Redis HA (#75) needs the same resource type for an unrelated reason, so both features converge on one replication group named `<cluster_name>-redis-rg`. `replication_group_id` forces replacement, so picking a neutral name before release avoids costing early adopters their queue a second time later.

PR #68, opened 2026-08-04
fix: size default HPA maxima to schedulable node capacity

### LINKED ISSUE #51: Default HPA max replicas exceed schedulable capacity of the default node group
## Problem

Default HPA maxima are 20 (main) and 50 (webhook processor). The default node group is capped at `node_max = 6` t3.xlarge instances, which is 24 vCPU total.

At default resource requests, 20 main pods (plus task runner sidecars) alone need more CPU than the whole cluster can ever provide. The HPA can scale to replica counts that are unschedulable by construction.

We hit this in practice: during a rollout with elevated CPU, the HPAs scaled toward their maxima and left 15 pods Pending with `Insufficient cpu` while the cluster autoscaler was already at `node_max`. The pending pods added churn and delayed rollout completion until we capped the HPAs manually.

## Suggested fix

One of:

- Lower the default HPA maxima to values the default node group can schedule (roughly 6 main / 8 webhook at default requests).
- Add a precondition or validation that warns when `hpa_max_replicas * cpu_request` exceeds `node_max * instance vCPU`.
- At minimum, document that these three variable groups (HPA maxima, resource requests, node group size) must be sized together.

buddy-n8n:
Resolves #51.

The default HPA maxima were unschedulable by construction. 20 main pods plus their task runner sidecars request 24,000m CPU on their own, which is more than the entire default node group (`node_max = 6` × t3.xlarge = 24 vCPU) can ever provide, before the worker and webhook ceilings are counted at all.

It survived this long because nothing fails loudly. During a rollout with elevated CPU the HPAs scale toward their maxima, the Cluster Autoscaler hits `node_max`, and the pods that don't fit sit `Pending` with `Insufficient cpu` while the surging ReplicaSet competes for the same exhausted CPU. #51 reports 15 such pods and a rollout that only finished once the HPAs were capped by hand.

## What changed

- `n8n_main_hpa_max_replicas` 20 → 6 and `n8n_webhook_hpa_max_replicas` 50 → 8. At the new defaults the three pod families request 16,600m against roughly 21,720m schedulable.
- A plan-time `check` that warns when the autoscaler ceilings, the per-pod CPU requests, and `node_max` × `node_instance_type` are out of step. Nothing in Kubernetes couples those three, so nothing tells you until pods are already `Pending`.
- New README section, "Sizing autoscaling against node capacity", plus a `docs/troubleshooting.md` entry for the symptom side (`Pending` / `Insufficient cpu` with the node group already at its maximum).
- `examples/medium` and `examples/large` pin the main pod range (3/24 and 6/60) against their own node groups instead of inheriting a starter default. The four small-sizing examples set no node group values at all, so they run the default node group these new defaults are sized for, and inheriting is correct there.
- A `spec.replicas` fix, see question 1 below.

### Why a `check` and not a `validation`

The model is an approximation, and staging higher ceilings before you raise `node_max` is a legitimate thing to do, so blocking either would be wrong.

vCPU is derived from the instance size suffix rather than read from `data.aws_ec2_instance_type`. A `check` whose condition is unknown at plan is a hard error rather than a warning, and a data source read is exactly what Terraform can defer to apply, so the precise version could turn an advisory hint into a broken plan. The suffix ladder is exact for 995 of the 1,150 instance types EC2 offers in eu-west-1. Bare-metal sizes silence the check entirely rather than warning off a guess, 48 sizes are over-counted (which costs a warning rather than raising a false one), and 3 legacy types are under-counted so the check can warn a hair early on those.

## Verification

Verified on a live 3-node cluster in us-east-1 and then destroyed, not checked by inspection. Full evidence is in the CHANGELOG entries. TLDR:

- Reproduced #51 exactly: 21 pods `Pending` with `0/3 nodes are available: 3 Insufficient cpu`, Cluster Autoscaler pinned at `node_max`.
- Measured every supply-side constant against the running cluster, which corrected three of them. The model predicted 10,500m available and the scheduler placed 10,200m. The 300m residue is bin-packing fragmentation, which no request-based model can represent, so read the warning as a bound and not a prediction.
- Reproduced and then confirmed the fix for the `spec.replicas` defect through a pod-churning `helm upgrade`.

CI was run locally on the pinned versions rather than my newer local toolchain (Terraform 1.9.8, terraform-docs v0.22.0, tflint v0.53.0): `fmt`, `terraform-docs --output-check` × 7, `validate` × 7, `test` × 7 (118 module tests plus every example), `tflint` × 7, `checkov`. All green. checkov's 28 findings are pre-existing and sit in files this PR doesn't touch.

## Upgrade impact

Both HPAs update in place with no resource replacement. Two cases need a caller's attention, both written up in `CHANGELOG.md`:

1. A cluster currently running more than 6 main or 8 webhook pods gets scaled back to the new ceiling on the next apply.
2. An explicit floor above the new ceiling fails the plan. `n8n_main_hpa_min_replicas = 8` or `n8n_webhook_hpa_min_replicas = 10` was valid against the old ceilings and isn't against the new ones. That's a caller who changes nothing and gets a plan error, so it's called out explicitly in the upgrade note.

## Two open questions

**1. Should the `spec.replicas` fix be split into its own PR?**

It's an independent defect that this work happened to surface, and it has its own `### Fixed` entry. The chart renders `spec.replicas` unconditionally on all three deployments (`multiMain.replicas`, `queueMode.workerReplicaCount`, `webhookProcessor.replicaCount`), whether or not an HPA or a KEDA ScaledObject also owns the field. The module passed a constant `2`, so every `helm upgrade` scaled main, worker and webhook down to 2 before their autoscalers climbed back.

Worth being precise about what the fix does, because cubic caught me overstating it: wiring each to its autoscaler's floor makes Helm's write a no-op *at rest*, and bounds the drop at the floor under load. It doesn't preserve an active scale-up. A webhook deployment at 80 now drops to 30 on `examples/large` instead of to 2, which is a real improvement and not a cure. Bounding it at the floor is the ceiling of what a chart consumer can do, since the field is rendered unconditionally and no value omits it, and reading the live replica count back into the plan would make every plan depend on cluster state. Eliminating it needs the chart to guard `spec.replicas` on whether an autoscaler owns the deployment. Happy to file that upstream against `n8n-helm-chart` if we agree it's the right ask.

For splitting: a) it isn't what #51 asked for, b) it's separately reviewable, c) it'd be separately revertable. Against: the warm floors this PR introduces for `examples/medium` and `examples/large` are unsafe without it, so shipping the raised floors while every `helm upgrade` still collapses them to 2 would make the regression worse, not better. If we do separate them, the `spec.replicas` fix should land first.

Splitting it back out is mechanical: three lines in `n8n.tf`, its CHANGELOG entry, and the floor wording in three variable descriptions. Happy to do it either way.

**2. The floor-above-ceiling validations went beyond the original ask, but they turned out to be load-bearing.**

Three `validation` blocks reject `min > max` for main, webhook and worker. #51 didn't ask for these, and they're the only hard failure this PR adds.

I originally flagged them as removable. They aren't. Lowering the two default ceilings is precisely what makes `min > max` reachable for a caller who changes nothing, which is upgrade case 2 above. Without them, that caller's plan succeeds and the apply fails partway through, with the Kubernetes API rejecting the HPA instead of Terraform naming the two inputs and their values.

Worth a second pair of eyes on the choice of a hard `validation` here while the capacity model next door is only a `check`. The distinction is deliberate: `min > max` is invariably rejected by Kubernetes and KEDA, whereas a ceiling above current node capacity is something a caller might legitimately be staging.

Thoughts?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,6 +175,28 @@ file. Use `command = plan` unless you specifically need apply semantics.
   intended to order a caller's resources, return the **resource attribute**
   rather than the variable, and treat "did the graph actually serialise these"
   as a question only a live apply answers.
+- **A `run` block cannot express "argument passed as `null`".** In a `variables`
+  block, `x = null` means *unset*: the variable takes its default and no error is
+  raised. A real caller writing `x = null` in a `module` block is different:
+  null **propagates into the module** rather than falling back to the default,
+  unless the variable declares `nullable = false`. Measured on 1.9.8, one module
+  with `default = 6` invoked three ways: no argument sees `6`; `n = null` on a
+  nullable variable sees `null`; `n = null` on a `nullable = false` variable sees
+  `6`, silently, with no error. Only the middle case can break anything, and it
+  is the only one a `run` block cannot reproduce.
+
+  This is not academic. It is how the `nullable = false` gap on the autoscaling
+  inputs was found (see `scaling.tf`): Terraform evaluates a `check` block's
+  `error_message` alongside its condition rather than lazily on failure, so a
+  propagated null reached the message's interpolations and aborted the plan with
+  `Invalid template interpolation value`, from a block whose entire purpose is
+  to warn *without* failing. `expect_failures` cannot assert it; the attempt
+  fails with "Missing expected failure".
+
+  So: for any input with a default where null is not meaningful, declare
+  `nullable = false` rather than trusting a test to catch it. Null-handling is a
+  question only a real `module` block answers, which means a live plan against a
+  caller configuration rather than the mocked suite.
 
 #### `check` conditions must short-circuit on Terraform 1.9
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -398,6 +398,17 @@ this project adheres to the stability contract in
   editor and REST API users, not a tier's executions/day. Resolves
   [#51](https://github.com/n8n-io/terraform-aws-n8n/issues/51).
 
+- `node_instance_type`, `node_max`, `n8n_task_runners_enabled`, and the six
+  autoscaler floor and ceiling inputs now declare `nullable = false`. A caller
+  passing an explicit `null` for one of these previously propagated it into
+  the module instead of falling back to the default, and once the capacity
+  check existed that null aborted the plan from inside the check's
+  interpolations (see the AGENTS.md note on null and `nullable = false`).
+  With `nullable = false`, Terraform substitutes the declared default at the
+  variable boundary. Only callers writing a literal `null` for one of these
+  nine inputs observe any change: they now get the default silently rather
+  than an error partway through the plan.
+
 - `aws_route53_record.cert_validation` keys its `for_each` off
   `local.acm_domain_names` instead of the certificate's computed
   `domain_validation_options`, and selects each record's values by matching
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,51 @@ this project adheres to the stability contract in
   from README.md and docs/troubleshooting.md, and the split-ingress example's
   `mcp_single_replica` workaround variable was removed with them.
 
+- Every `helm upgrade` scaled the main, worker and webhook deployments down to 2
+  replicas before their autoscalers climbed back. The chart renders
+  `spec.replicas` unconditionally on all three deployments, from
+  `multiMain.replicas`, `queueMode.workerReplicaCount` and
+  `webhookProcessor.replicaCount`, with no regard for whether an HPA or a KEDA
+  ScaledObject also owns the field. The module passed a constant `2` for all
+  three, so Helm and the autoscaler disagreed on every apply.
+
+  It reset any scale-up, not only a configured floor: a deployment the HPA had
+  taken to 8 under load went back to 2 and had to climb again. It cost most where
+  the floor was highest, since `examples/large` runs a webhook floor of 30 and a
+  worker floor of 20, both of which collapsed to 2 on each apply and then had to
+  be re-scaled while traffic was already arriving. Same class of defect as
+  [#51](https://github.com/n8n-io/terraform-aws-n8n/issues/51): pod count driven
+  by two sources that disagree.
+
+  All three are now wired to their autoscaler's floor
+  (`n8n_main_hpa_min_replicas`, `n8n_worker_keda_min_replicas`,
+  `n8n_webhook_hpa_min_replicas`), making Helm's write a no-op. Deployments on
+  the default floors see no change, since those already matched the constant.
+
+  Verified on a live cluster rather than by inspection, because no plan-time test
+  can observe a `helm upgrade`. A 3-node cluster was applied from the previous
+  version with a webhook floor of 5 and a worker floor of 3, then given one
+  unrelated config change to trigger an upgrade. Sampling `spec.replicas` every
+  two seconds through it:
+
+  ```
+     5 samples  webhook=5 worker=3   <- steady state at the configured floors
+     2 samples  webhook=2 worker=2   <- Helm re-asserts spec.replicas=2
+    41 samples  webhook=5 worker=3   <- HPA and KEDA climb back
+  ```
+
+  Webhook capacity dropped 60% and worker 33% mid-upgrade, and `helm get manifest`
+  confirmed the release's stored desired state read `replicas: 2` for all three
+  while the cluster ran 5 and 3. Repointing at this version planned `Plan: 0 to
+  add, 1 to change, 0 to destroy`, an in-place `helm_release` update with no
+  replacement of the ALB, RDS instance, KMS key or any PVC, and moved the stored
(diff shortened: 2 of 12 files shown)


buddy-n8n:
All four findings were valid, fixed in 579ff39. Comments and docs only, no logic change.

On the P2, you are right and I had overstated the fix. Confirmed against the pinned chart (1.10.0) that there is no autoscaler guard on any of the three templates:

```
deployment-main.yaml:15:              replicas: {{ ternary .Values.multiMain.replicas .Values.replicaCount .Values.multiMain.enabled }}
deployment-worker.yaml:15:            replicas: {{ .Values.queueMode.workerReplicaCount }}
deployment-webhook-processor.yaml:15:  replicas: {{ .Values.webhookProcessor.replicaCount }}
```

So wiring `spec.replicas` to the floor makes Helm's write a no-op at rest and bounds the drop at the floor under load, but it does not preserve an active scale-up. On `examples/large` a webhook deployment at 80 now drops to 30 rather than to 2. Real improvement, not a cure. My live verification only sampled deployments resting at their floors, so it could not have caught this; `CHANGELOG.md` now says so explicitly.

I did not go further in this PR because bounding at the floor is as far as a chart consumer can get: the field is rendered unconditionally so no value omits it, and reading the live replica count back into the plan would make every plan depend on current cluster state. Your suggested fix is the right one and it belongs upstream in `n8n-helm-chart`, guarding `spec.replicas` on whether an autoscaler owns the deployment. Flagged in the PR description.

The stale README table (finding 2) was the most useful catch. `examples/small` sets no HPA inputs, so it inherits the default this PR lowered to 8, and the table was overstating small-tier webhook capacity by 6x. That table is not in this PR's diff, which is exactly why a diff-scoped review missed it. Corrected to `2 / 8` and added the missing main pod row.

jrx (commented):
Reviewed on top of `2f0de4c` with the repo's full local loop: `terraform fmt -check`, `validate`, `terraform test` (118 passed), `tflint`, `terraform-docs --output-check`, plus validate and the test suites for all six examples. Everything is green. Spot-checked the capacity model: the kubelet reserve curve resolves 4 vCPU to 80m reserved, matching the stated 3,920m allocatable for t3.xlarge, and the check block follows the repo's 1.9 `guard ? body : true` rule.

Three minor points, none blocking:

1. **`scaling.tf:177` and `scaling.tf:195`** - `node_daemonset_cpu_millis = 180` and `cluster_addon_cpu_millis = 720` are measured from the chart and addon versions the module currently pins, but nothing ties them to those pins. A future bump of the AWS LBC, Cluster Autoscaler, KEDA, EBS CSI, or metrics-server versions can silently drift the model. Suggest a one-line comment on each constant naming the version inputs whose bumps should trigger re-verification, so the maintenance trigger is discoverable at the constant rather than in git history.

2. **`variables.tf`** - the 11 new `nullable = false` declarations are an interface semantics change that reaches beyond the capacity check: a caller passing an explicit `null` previously propagated it into the module, and now silently receives the default instead. That is the better behavior, but it deserves its own bullet under `### Changed` in CHANGELOG.md rather than living only inside the capacity-check narrative, since it also changes what callers of inputs like `node_instance_type` observe.

3. **`tests/defaults.tftest.hcl:33`** - stray extra blank line added inside the `mock_provider "aws"` block. Cosmetic only.

For context: this supersedes a parallel branch implementing the same default reduction (20 to 6, 50 to 8, identical capacity arithmetic). That branch is being retired in favor of this PR, since this one additionally fixes the `spec.replicas` scale-down defect and adds the floor/ceiling validation and the advisory capacity check.

jrx (approved):
Approving after full validation of the rebased branch plus end-to-end verification on a live cluster (EKS `pr68`, us-east-1, deployed from `examples/small`).

**Static and test validation** (module root and all six examples): `terraform fmt`, `validate`, `terraform test` (118 passed), `tflint`, `terraform-docs --output-check` all green.

**Live verification, change by change:**

- Default maxima deployed correctly: main HPA 2/6, webhook HPA 2/8, KEDA 1/10.
- Capacity check warns without failing on real providers: plan with the old ceiling of 20 produced the warning with correct arithmetic (33,400m requested vs 21,720m schedulable) and still planned `0 add / 1 change / 0 destroy`.
- Floor-above-ceiling validation fails the plan before anything reaches AWS, naming both inputs.
- `spec.replicas` floor wiring: deployments rest at 2/2/1 (worker at its KEDA floor rather than the old constant 2), confirmed in both the stored Helm manifest and live state.
- Replica stability through a real pod-churning `helm upgrade` (image `stable` to `2.33.3`): deployment events show only the rolling-update surge pattern; main and webhook aggregate never dropped below 2 and worker never below 1. No Helm-driven collapse.
- Floors do not impede scaling: KEDA scaled workers 1 to 10 under smoke-test load and back down cleanly.
- Full smoke test suite with `LOAD_TEST=true`: 32 passed, 0 failed, 0 warnings, 0 skipped.

Reminder for the release: changed defaults, so this ships as 0.3.0 per the stability contract.

PR #69, opened 2026-08-04
feat: support execution data in S3 via n8n_execution_data_storage_mode

### LINKED ISSUE #47: feat: support execution data in S3 (N8N_EXECUTION_DATA_STORAGE_MODE, n8n >= 2.27)
### What problem are you trying to solve?

n8n v2.27.0 added a second S3 offload mode: execution data itself can now be stored in S3 instead of PostgreSQL via `N8N_EXECUTION_DATA_STORAGE_MODE=s3` (shipped in [n8n#32226](https://github.com/n8n-io/n8n/pull/32226), documented in [n8n-docs#4808](https://github.com/n8n-io/n8n-docs/pull/4808)). Each new execution is written as a `bundle.json` under `workflows/{workflowId}/executions/{executionId}/execution_data/`.

Execution-data writes are typically the dominant write load on the n8n database at volume, so this mode directly relieves RDS pressure in exactly the queue-mode architecture this module deploys. The module already provisions everything the feature needs: the shared S3 bucket, the IAM policy (`s3:GetObject/PutObject/DeleteObject/ListBucket`), and the `N8N_EXTERNAL_STORAGE_S3_*` connection wiring used by binary data mode. What's missing is a supported way to turn the new mode on: `N8N_EXECUTION_DATA_STORAGE_MODE` is not rendered by the module today, and the env-guard in `locals.tf` doesn't list it, so callers can't cleanly opt in.

### Proposed solution

1. **New input** `n8n_execution_data_storage_mode` (string, default `"database"`, validated against `["database", "s3"]`), or alternatively a boolean `n8n_execution_data_s3_enabled` (default `false`). When set to S3, the module renders `N8N_EXECUTION_DATA_STORAGE_MODE=s3` on main, workers, and webhook processors, reusing the existing bucket and `N8N_EXTERNAL_STORAGE_S3_*` connection.
2. **Env-guard update:** add `N8N_EXECUTION_DATA_STORAGE_MODE` to the module-managed exact-match list in `locals.tf` once the module owns it.
3. **Lifecycle-rule scoping (correction, 2026-07-16):** an earlier version of this issue proposed a lifecycle input for execution data. That was wrong: per the [external storage docs](https://docs.n8n.io/hosting/scaling/external-storage/), n8n prunes execution data in S3 **itself** (the prune hard-delete path removes the S3 bundle together with the DB record), and the docs explicitly warn against lifecycle rules on execution data because they could delete bundles n8n still references. The actual module concern is the inverse: binary data **requires** a lifecycle rule (n8n delegates binary pruning to S3), and the module's single shared bucket will hold both data types once this mode lands. Any lifecycle configuration the module creates or documents should therefore be **scoped to the `binary_data/` object paths**, never bucket-wide, so it can't touch `execution_data/` bundles.
4. **Docs note:** requires n8n >= 2.27 (`n8n_image_tag`); licensing follows binary data in S3; no backfill — existing executions stay in the database, only new executions are written to S3.

### Alternatives considered

- Injecting the env var through the module's extra-env passthrough: blocked (correctly) by the env-guard, and it wouldn't cover the lifecycle-policy half.
- Managing a separate bucket outside the module: loses the module's IAM/IRSA wiring and splits state for no benefit; the feature is designed to share the binary-data S3 connection.

buddy-n8n:
Adds `n8n_execution_data_storage_mode` so execution data can go to the S3 bucket the module already creates for binary data, instead of PostgreSQL. n8n 2.27 added this as a second S3 offload mode (`N8N_EXECUTION_DATA_STORAGE_MODE`, n8n-io/n8n#32226), and execution-data writes are usually the dominant write load on the database at volume, so this is the main lever for relieving RDS pressure in the queue-mode topology this module deploys.

Default is `"database"`, which emits no env var at all, so existing deployments see a no-op diff. `"filesystem"` is rejected outright: pod filesystems are ephemeral and unshared here. There's a plan-time check that warns when the mode is `"s3"` while `n8n_image_tag` is pinned below 2.27, since the env var is silently ignored on older versions and execution data just keeps going to PostgreSQL.

Breaking change, worth calling out directly: `N8N_EXECUTION_DATA_STORAGE_MODE` is now reserved in the `n8n_extra_env` guard. The guard actually never covered this name (no prefix match, not on the exact-match list either), so anyone setting it through the escape hatch will hit a validation failure and need to move to the new variable. Puts this on a minor version bump, not a patch.

All six examples now expose the passthrough, all defaulting to `"database"`. The four small-sizing ones (`small`, `cloudflare`, `godaddy`, `split-ingress`) carry byte-identical wiring, since they're the same deployment differing only in the one thing each demonstrates, and they're also the ones with the least database headroom to absorb execution-data growth (`db.t3.small`, 50 GB gp2, 150 IOPS baseline) even though they see the least traffic. `large` runs Aurora I/O-Optimized with no IOPS ceiling, so it's the tier best equipped to just eat the volume instead.

Two hazards documented in the README. First, the shared bucket can't get an S3 lifecycle rule that touches both object types: binary data is pruned only by S3 (n8n delegates it), execution data is pruned by n8n itself, and a lifecycle rule can delete bundles n8n still references. Can't split them by prefix either, S3 lifecycle filters are literal-prefix only and both layouts share `workflows/{wf}/executions/{exec}/`. Second, durability: RDS gets 7 days of backups and PITR by default, the bucket has no versioning, no backups, and `force_destroy = true`, so a `terraform destroy` in s3 mode takes execution history with it, permanently.

Verified live against n8n 2.33.1 on the pinned chart (1.10.0): env var reaches main, worker, and webhook-processor pods (spec and runtime), flipping the mode is a single in-place change, key layout is `workflows/{wf}/executions/{exec}/execution_data/bundle.json`, existing Pod Identity policy needs nothing extra, read path works, `execution_entity.storedAt` tracked the mode correctly across four flips (s3 mode writes no `execution_data` row at all), the scheduled pruner removed bundles on its own timer and left `binary_data` orphaned (normal steady state), `force_destroy` took a bucket with 8 objects in it clean, and the license gate fails with the correct entitlement named: `feat:executionDataS3`, distinct from `feat:binaryDataS3` which the always-on binary offload uses.

Test plan:
- [x] `terraform test` on terraform 1.9.8 (CI's pin, not local 1.15.8): root 108, small/medium 2, large 15, cloudflare 6, godaddy 7, split-ingress 14, all green
- [x] `terraform fmt -check -recursive`
- [x] `terraform-docs --output-check` on all 7 dirs, pinned to v0.22.0 (CI's version, not local 0.24.0)
- [x] live apply against a real EKS cluster: mode flip both directions, pruning, destroy, license gate

Resolves #47


buddy-n8n:
Both checked out, one real, one not quite.

**Violation 1 (split-ingress duplicate `n8n_image_tag`)**: confirmed. The example already had `n8n_image_tag = "2.33.1"` and my new block added a second, disagreeing assignment. Fixed by dropping the duplicate and pointing the comment at the existing pin instead.

**Violation 2 (medium test contradicting the adjacent note)**: the added test itself is fine, actually reproduced this rather than taking it on faith. Added a throwaway `n8n_image_tag_rejects_whitespace` run to both `small` and `medium` on terraform 1.9.8 (the version CI pins) and it passes clean, no dns.tf for_each collision. That comment predates this PR (it's on `main` already), so it was already stale, this PR just sat a new test next to it. Replaced the stale note in both files with the real `n8n_image_tag_defaults_to_null` / `n8n_image_tag_rejects_whitespace` tests that `large`, `cloudflare`, and `godaddy` already carry, so all six examples are now consistent.

Pushed as 55ba791. Full suite still green: root 108, small/medium 4, large 15, cloudflare 6, godaddy 7, split-ingress 14.

jrx (approved):
Approving after a full code review and live validation of this branch.

## Code review

- Rebased cleanly onto main, single commit, original author preserved.
- Input follows the repo conventions: description, type, validation, nullable = false, default matching n8n's own default so existing deployments see a zero diff.
- The plan-time version check correctly uses nested ternaries for the Terraform 1.9 short-circuit limitation.
- Reserving N8N_EXECUTION_DATA_STORAGE_MODE in the extra_env guard is documented in the CHANGELOG as a minor-version boundary with a clear migration path.
- Full local loop green: fmt, validate, terraform test (root and all six examples), tflint, terraform-docs.

## Live validation (examples/small, us-east-1, cluster pr69)

- Phase 1 apply with defaults: 75 added, 0 changed, 0 destroyed. Confirmed the env var is absent from every container in database mode (no-op guarantee).
- Phase 2 flip to s3 with n8n_image_tag = 2.33.4: exactly one in-place helm_release change. Env var landed on main, worker, and webhook-processor pods; atomic upgrade succeeded, proving the feat:executionDataS3 license gate passed (verified beforehand via n8n license:info).
- Smoke test: 32 passed, 0 failed, including queue-mode execution and KEDA worker scaling 1 to 4.
- End-to-end S3 lifecycle verified on live infra:
  - Write: execution produced workflows/{wf}/executions/102/execution_data/bundle.json (2.7 KB), exactly the documented key layout.
  - Read-back: API returned the full execution data with the payload living only in S3.
  - Prune: deleting the workflow hard-deleted the execution and n8n removed the S3 bundle itself, bucket back to zero objects. This also confirms the README's rationale for shipping no lifecycle configuration.

All 24 CI checks green. Ready to merge.

PR #70, opened 2026-08-04
fix: give KEDA's Redis triggers TLS and the AUTH token

### LINKED ISSUE #66: KEDA worker autoscaler loses Redis auth when redis_transit_encryption_enabled is set

buddy-n8n:
Resolves #66. **Stacked on #67** (issue #41), which introduces the AUTH token, the Kubernetes secret and `local.redis_host` this depends on. Base is `feat/issue-41-redis-transit-encryption`, so merge #67 first.

## The problem

With `redis_transit_encryption_enabled = true`, the worker ScaledObject's Redis triggers carried neither TLS nor credentials. KEDA could not read queue depth, so workers held a static replica count instead of tracking backlog. Nothing crashed and n8n was unaffected, which is what made it easy to miss.

## The fix

Both triggers merge in two metadata keys when the flag is on:

```hcl
enableTLS       = "true"
passwordFromEnv = "QUEUE_BULL_REDIS_PASSWORD"
```

**TLS is the half that has to land.** The live evidence on #66 shows the first failure is `connection to redis failed: i/o timeout`, not `NOAUTH`: KEDA opens a plaintext connection to a TLS-only endpoint and hangs before authentication is ever attempted. Adding credentials without `enableTLS` would have read as no fix at all.

**The token is never written into the ScaledObject.** `passwordFromEnv` names an environment variable, and KEDA resolves it against the scale target's `containers[0]`, following the `secretKeyRef` the chart already sets there. The chain, confirmed in KEDA source and then on the live object:

1. The redis scaler tags its password field `keda:"name=password, order=triggerMetadata;resolvedEnv;authParams"`.
2. `resolvedEnv` is implemented as `TriggerMetadata["<name>FromEnv"]` giving an env var name, then `ResolvedEnv[thatName]`.
3. `ResolvedEnv` is built by `resolveEnv`, which handles `ValueFrom.SecretKeyRef`, gated on `accessSecrets`.
4. `accessSecrets` is false only under `KEDA_RESTRICT_SECRET_ACCESS=true`; the KEDA chart defaults both `restrict.secret` values to `false` and `keda.tf` does not override them.
5. `containers[0]` on the worker Deployment is `n8n-worker` (`task-runner` is a later sidecar), which is exactly where the chart mounts the token.

This is why no `TriggerAuthentication` resource is needed, which matters: it keeps `kubernetes_manifest` out of this module. That resource reads the target CRD's schema from a live cluster **at plan time**, so on a from-scratch apply the plan would fail before `helm_release.keda` had created the KEDA CRDs, forcing a two-stage apply and breaking the module's single-`apply` promise.

No chart change is required either. `scaledobject-worker.yaml` renders arbitrary metadata keys via `range $key, $value := .metadata`.

## Scope

Worker-only. The open question on #66 about the webhook-processor ScaledObject resolves to no: its template is guarded on `.Values.keda.webhookProcessor.enabled`, which defaults to `false` and which this module never sets. Webhook scaling is the CPU HPA in `scaling.tf`.

## Verification

Full results in the deploy harness notes. Live cluster in us-east-1, KEDA **2.20.2** (what the unpinned chart resolved to), n8n 2.33.1, flag on from the first apply.

| | before (#41 run) | after |
|---|---|---|
| ScaledObject | `READY=False  ACTIVE=Unknown` | `READY=True  ACTIVE=False` |
| HPA targets | `<unknown>/5 (avg), <unknown>/5 (avg)` | `0/2 (avg), 0/2 (avg)` |
| KEDA operator | `connection to redis failed: i/o timeout` | no errors |

A numeric `0` rather than `<unknown>` is the claim: KEDA completed a metric read, which needs both the handshake and authentication to succeed. `grep -icE "i/o timeout|noauth|wrongpass"` over the operator log returned **0** across the whole run, including under load.

Scale-up and scale-down, at `min = 2`, `max = 6`, `jobs_per_replica = 2`:

```
16:41:07  backlog -> 20
16:42:01  workers 6/6      (up from 2, ~55s; demand was ceil(20/2) = 10, capped at max)
16:44:58  backlog cleared
16:50:00  workers 2/2
```

Scale-down took about 5 minutes, consistent with the HPA's default 300s scale-down stabilization window on top of KEDA's 60s `cooldownPeriod`.

The 64-character AUTH token appears **0 times** in the rendered ScaledObject and HPA YAML.

This exercises the entire path the fix touches: handshake, AUTH, `LLEN` read, metric served to the HPA, and the HPA moving replicas both directions. It does not re-demonstrate that n8n itself produces those entries over the encrypted connection. That was already settled in the #41 run, where a real workflow was enqueued by a main and executed by a worker across the same TLS+AUTH Redis. The two runs together cover the full path.

## Docs

This **retracts a documented limitation**, so it is undone everywhere #41 wrote it: the `redis_transit_encryption_enabled` description, the README section (now describing working behaviour plus the `KEDA_RESTRICT_SECRET_ACCESS` caveat), and the CHANGELOG.

#66 asks for a `### Fixed` bullet rather than rewriting the `### Added` one. That assumed #41 had shipped. It has not, so both land in the same unreleased version and a "Fixed" entry would describe a bug no user ever saw. The `### Added` entry is amended instead. Happy to split it if you would rather have the separate entry.

## Gates

`terraform fmt` clean, `validate` clean, `tflint` clean, `terraform-docs` clean, **107 tests pass** (2 new: trigger metadata is empty by default, and carries TLS plus the env reference and no literal password when enabled). Checkov findings are pre-existing and unchanged; this adds no AWS resources.

The verification cluster has been destroyed.


buddy-n8n:
Note for context: cubic pushed a commit directly to #67 (`d9612c6`) that fixed this same KEDA/Redis auth gap by adding a `kubernetes_manifest` `TriggerAuthentication` resource. I reverted that commit on #67 (`c277d27`), since this PR fixes the issue without requiring a CRD, using `passwordFromEnv` instead.

buddy-n8n:
Superseded by #76, which carries this fix.

The KEDA implementation here is the one that shipped: `local.keda_redis_auth_metadata` merged into both trigger metadata blocks, `enableTLS` plus `passwordFromEnv`, token by env-var reference so it never lands in the ScaledObject. The live verification recorded on this branch (KEDA 2.20.2, HPA reading a numeric value rather than `<unknown>`, 2 to 6 workers in ~55s) carried forward into #76's description.

What changed is where it lives. This branch and #67 each defined their own `aws_elasticache_replication_group.n8n`, and so does the Redis HA work in #75. Terraform allows one, so all three had to collapse onto a single resource whose attributes vary by both toggles. #76 does that, and closes both #41 and #66.

One thing found while merging them: `auth_token_update_strategy` cannot be set unconditionally. The provider rejects it when there is no token (`"auth_token_update_strategy": "auth_token" must be specified`), which breaks the HA-only plan. Neither `terraform validate` nor `terraform test` catches it, because the test framework's mocked provider skips the real provider's argument validation.

PR #72, opened 2026-08-04
fix(eks): ignore desired_size drift on node group

### LINKED ISSUE #50: aws_eks_node_group desired_size fights the cluster autoscaler
## Problem

The module tags the node group for cluster autoscaler ownership (`k8s.io/cluster-autoscaler/enabled = true`), but `desired_size` is not excluded from drift detection. When the autoscaler scales the node group, every subsequent plan wants to reset `desired_size` back to `var.node_desired`.

Applying that change terminates live nodes. In our case (module 0.2.0), an apply during an HPA scale-up event reset `desired_size` from 6 to 3, drained 3 nodes mid-rollout, evicted n8n pods, and amplified the scaling event it was applied during.

## Suggested fix

Ignore `desired_size` drift, which is the standard pattern for autoscaler-managed node groups:

```hcl
resource "aws_eks_node_group" "n8n" {
  # ...
  lifecycle {
    ignore_changes = [scaling_config[0].desired_size]
  }
}
```

`node_desired` then only sets the initial size, which the variable description should mention.

buddy-n8n:
## Summary

- The node group is tagged for Cluster Autoscaler auto-discovery, but Terraform still tracked `scaling_config.desired_size` as authoritative. Once the autoscaler moved the live count away from `var.node_desired`, every plan proposed resetting it back down, and applying that drains live nodes.
- Adds `scaling_config[0].desired_size` to `aws_eks_node_group.n8n`'s `lifecycle.ignore_changes`, the standard pattern for autoscaler-managed node groups.
- Updates `node_desired`'s description to clarify it only sets the size at creation.

Fixes #50

## Test plan

- [x] `terraform validate`
- [x] `terraform fmt -check`
- [x] `terraform test` (101 passed, 0 failed, including the existing `scaling_config` assertions in `tests/defaults.tftest.hcl`)
- [x] `terraform-docs .` regenerated `README.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ this project adheres to the stability contract in
   `desired_size` as authoritative. Once the autoscaler moved the live count
   away from `var.node_desired`, every subsequent plan proposed resetting it
   back down, and applying that drains live nodes: reported in
-  [#50](https://github.com/n8n-io/terraform-aws-n8n/issues/50), an apply
+  [#50](https://github.com/n8n-io/terraform-aws-n8n/issues/50) on module
+  0.2.0, an apply
   during an HPA scale-up event reset `desired_size` from 6 to 3, draining 3
   nodes mid-rollout and evicting n8n pods. `scaling_config[0].desired_size`
   is now in the node group's `lifecycle.ignore_changes`, the standard pattern
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ this project adheres to the stability contract in
 
 ### Fixed
 
+- `aws_eks_node_group.n8n` no longer fights the Cluster Autoscaler over
+  `desired_size`. The node group is tagged for autoscaler auto-discovery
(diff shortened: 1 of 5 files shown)


jrx (approved):
Reviewed against Terraform Registry and repo AGENTS.md standards.

- Fix is the standard pattern for autoscaler-managed node groups: `lifecycle { ignore_changes = [scaling_config[0].desired_size] }` on `aws_eks_node_group.n8n`. No interface change, no state migration risk; `node_desired` description now correctly documents create-time-only semantics.
- Branch is cleanly rebased on current `main`, linear history.
- Verified locally: `terraform fmt -check -recursive`, `terraform validate`, `terraform test` (101 passed, 0 failed), `tflint` clean, `terraform-docs --output-check .` in sync, plus validate + test across all six examples.
- Follow-up commit 58a1765 adds the AGENTS.md-prescribed test-file comment (why the drift suppression cannot be asserted under mocked plan-time tests and how to verify it live) and notes the affected module version in the CHANGELOG.

Approving. This supersedes the parallel implementation on `jrx/ignore-node-desired-size-drift`, which has been deleted.

PR #73, opened 2026-08-04
fix: warn and expose HPA behavior for webhook reinstall_missing_packages spikes

### LINKED ISSUE #52: Resource defaults too low when n8n_reinstall_missing_packages is enabled
## Problem

With `n8n_reinstall_missing_packages = true`, every pod runs npm installs at boot. n8n also rebroadcasts installs to all pods via pubsub, so during a rolling restart every pod installs repeatedly. Against the webhook processor defaults (300m CPU request, 800m limit, 1Gi memory limit) this causes two failure modes we hit in production on module 0.2.0:

1. **CPU:** installs burn 800-1000m per pod. That reads as 200-300% of the 300m request, so the CPU-based HPA scales up on every rollout. Each new pod boots, installs, and broadcasts, which keeps utilization above target. This feeds back into scale-up until capacity runs out.
2. **Memory:** concurrent installs plus the n8n baseline exceed 1Gi. Pods get OOMKilled (exit 137) mid-install, restart, reinstall, and broadcast again. This is a self-feeding crash loop, and interrupted installs left corrupted package directories (`ENOTEMPTY`, `tar: invalid magic`).

Stable values for us: webhook 800m/1500m CPU and 1Gi/2Gi memory.

## Suggested fix

- Raise the webhook processor defaults, or document that `n8n_reinstall_missing_packages = true` requires larger requests and limits.
- Optionally expose the HPA `behavior` block (scale-up stabilization window) so short boot spikes don't trigger scaling. The `kubernetes_horizontal_pod_autoscaler_v2` resource in `scaling.tf` currently hardcodes no behavior.

buddy-n8n:
## Summary

With `n8n_reinstall_missing_packages = true`, every pod runs npm installs at boot, and n8n rebroadcasts installs to all pods via pubsub, so a rolling restart makes every webhook pod install repeatedly at once. Against the webhook processor's default resources this produces two production failure modes: CPU spikes read as 200-300% of the request and drive the CPU-based HPA into a scale-up-on-every-rollout loop, and concurrent installs plus the n8n baseline can exceed the memory limit and OOMKill pods mid-install into a crash loop.

- Adds a plan-time `check` block (`webhook_resources_sized_for_reinstall_missing_packages` in `scaling.tf`) that warns when `n8n_reinstall_missing_packages = true` and the webhook processor's CPU/memory requests/limits are below the values known to survive it in production. Advisory only, defaults unchanged, existing deployments are unaffected.
- Adds `n8n_webhook_hpa_scale_up_stabilization_window_seconds` (default `0`, matching the Kubernetes API's own default) to expose the previously-hardcoded `behavior.scaleUp` block on the webhook HPA, so a short boot-time CPU spike can be absorbed instead of immediately triggering a scale-up.
- Documents the tradeoff on `n8n_reinstall_missing_packages` and the four webhook resource variables, and adds a `docs/troubleshooting.md` section walking through both failure modes and the fix.

Fixes #52

## Test plan
- [x] `terraform fmt -check -recursive`
- [x] `terraform validate` (root + `examples/small`, `examples/medium`, `examples/large`)
- [x] `terraform test -verbose` (root: 107 passed; small: 1; medium: 1; large: 14 — all passed)
- [x] `tflint --format compact` (clean)
- [x] `terraform-docs --output-check .` (README in sync, using CI-pinned v0.22.0)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,17 @@ this project adheres to the stability contract in
   before anything is sent to the cluster. A floor equal to its ceiling stays
   valid, since that is how you pin a fixed-size deployment.
 
+- `n8n_webhook_hpa_scale_up_stabilization_window_seconds` input (default `0`,
+  matching the Kubernetes API's own default) exposes
+  `behavior.scaleUp.stabilizationWindowSeconds` on the module-managed
+  `kubernetes_horizontal_pod_autoscaler_v2.n8n_webhook`. Lets a caller absorb a
+  short boot-time CPU spike (for example from
+  `n8n_reinstall_missing_packages = true`) without an immediate scale-up. A new
+  `webhook_resources_sized_for_reinstall_missing_packages` plan-time check warns
+  when `n8n_reinstall_missing_packages = true` and the webhook processor's
+  CPU/memory requests and limits are below the values known to survive it in
+  production. See [issue #52](https://github.com/n8n-io/terraform-aws-n8n/issues/52)
+  and `docs/troubleshooting.md`.
 - `n8n_license_detach_floating_on_shutdown` input (default `false`) maps to
   `N8N_LICENSE_DETACH_FLOATING_ON_SHUTDOWN`, overriding n8n's own upstream
   default of `true`. In multi-main (the module default), the leader main
--- a/README.md
+++ b/README.md
@@ -183,8 +183,7 @@ Six runnable examples ship with the module: three sizing tiers (`small`, `medium
 | DB IOPS ceiling | 150 baseline / 3,000 burst | 3,000 baseline (gp3) | None — I/O-Optimized |
 | PgBouncer | No | No | Yes — 2 replicas |
 | Redis | cache.t3.medium | cache.r6g.large | cache.r6g.large |
-| Main pods min / max | 2 / 6 | 3 / 24 | 6 / 60 |
-| Webhook pods min / max | 2 / 8 | 5 / 50 | 30 / 80 |
+| Webhook pods min / max | 2 / 50 | 5 / 50 | 30 / 80 |
 | Worker pods min / max | 1 / 10 | 5 / 40 | 20 / 160 |
 | Worker concurrency | 10 | 20 | 40 |
 | Execution concurrency limit | 100 | 200 | 2,000 |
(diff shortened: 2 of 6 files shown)


buddy-n8n:
Reviewed cubic's two findings:

1. **Indentation at scaling.tf:88** — false positive. That indentation is exactly what `terraform fmt` produces for a `for ... in { ... } : k => v` expression (the inline map's closing brace outdents to align with `for`, not with the map's keys). Confirmed with `terraform fmt -check -diff scaling.tf` (no diff) and CI's `terraform fmt` check on the same commit was green.
2. **Decimal quantities silently bypassing the check** — valid, fixed in 3d08682. `"0.5"` CPU and `"0.5Gi"` memory now parse correctly instead of being treated as unreadable. Added a regression test (`webhook_resources_decimal_cpu_below_threshold_triggers_check_warning`).

Full suite re-run: 108 passed, 0 failed.

jrx (approved):
Approved after full validation of this branch, rebased onto current main (d6cde52).

**Static checks** (module root): fmt, validate, `terraform test` 125 passed / 0 failed, tflint, terraform-docs — all green.

**Live verification** (fresh `examples/small` deployment in us-east-1, 75 resources):
- HPA `behavior.scaleUp` deployed and confirmed via kubectl: `stabilizationWindowSeconds: 0`, `selectPolicy: Max`, default-replicating policies. HPA healthy (`ScalingActive=True`, cpu 0%/65%).
- Unmanaged `scaleDown` picks up the API server defaults with no plan drift (`terraform plan` → no changes).
- `webhook_resources_sized_for_reinstall_missing_packages` check verified against real state: warns with the full actionable message when `n8n_reinstall_missing_packages = true` at default resources, silent once sized to 800m/1500m + 1Gi/2Gi, silent on the default path.
- Stabilization override (0 → 300) plans as an in-place HPA update.
- Post-deploy smoke test: 31 passed / 0 failed, including an end-to-end queued workflow execution.

**Note:** the initial live apply surfaced one bug — `scale_up` without `select_policy` serializes as `selectPolicy: ""` and the Kubernetes API rejects the HPA at admission (invisible to the mocked test suite). Fixed in ac4e400 with an explicit `select_policy = "Max"` plus a plan-time regression assert; that commit and the rebase onto main were pushed to this branch.

PR #75, opened 2026-08-05
feat: optional Redis high availability and an external ElastiCache hook

### LINKED ISSUE #44: feat: optional Redis high availability (replication group with automatic failover)
## Problem

The module provisions Redis as a single-node `aws_elasticache_cluster` (`redis.tf`, `num_cache_nodes = 1`), unconditionally. Redis backs the Bull queue that distributes executions across workers and coordinates the multi-main leader, so a single node is a single point of failure: a node failure or AZ event stalls the queue and the multi-main coordination until ElastiCache replaces the node.

The current sizing decision validated `cache.r6g.large` for *memory headroom* only; failover / HA was never decided. Separately, the proposed cross-region HA/DR design assumes a `create_elasticache = false` caller hook (mirroring the existing `create_database = false` hook) so both regions can point at a shared, replication-capable Redis endpoint — but that hook does not exist yet. Today the Redis tier is neither internally HA nor externalizable.

## Proposal

Two related, opt-in additions (could ship as one issue or be split):

1. **Optional intra-region HA.** Add `redis_high_availability_enabled` (bool, default `false`). When `true`, provision an `aws_elasticache_replication_group` with a replica and `automatic_failover_enabled = true` (Multi-AZ) instead of the single-node cluster, keeping `redis_node_type`. Default preserves the current single-node topology.
2. **External Redis hook.** Add `create_elasticache` (bool, default `true`) plus `redis_host` / `redis_port` inputs, mirroring `create_database` / `db_host`. When `false`, skip module-managed Redis and wire n8n at the supplied endpoint. This is the hook the cross-region HA/DR design already presumes.

Moving from `aws_elasticache_cluster` to `aws_elasticache_replication_group` is a resource-type change, so migration needs care (`moved` blocks are not possible across resource types) — the issue should decide whether HA is a new-deployment-only path or whether a documented recreate is acceptable.

## Notes

- Both additions are opt-in with defaults that preserve current behavior (single node, module-managed).
- Follows the curated-surface contract (dedicated, validated, documented variables).
- Relates to the cross-region HA/DR design, which depends on the `create_elasticache = false` hook proposed here.
- Needs `variables.tf` entries with validation, wiring in `redis.tf` + `n8n.tf`, `.tftest.hcl` asserts, a CHANGELOG bullet, and a `terraform-docs` refresh. Migration/replacement semantics must be documented in the variable description and README.

buddy-n8n:
Closes #44.

Two opt-in additions to the Redis tier, both defaulting to exactly what the
module does today.

**`redis_high_availability_enabled`** provisions Redis as a two-node
`aws_elasticache_replication_group` (primary plus replica, `automatic_failover_enabled`,
`multi_az_enabled`) instead of the single-node `aws_elasticache_cluster`. Redis backs
both the Bull queue and multi-main leader election, so the default single node is a
single point of failure for both.

**`create_elasticache = false`** plus `redis_host` / `redis_port` mirrors the existing
`create_database` / `db_host` hook, and is what the cross-region HA/DR design presumes:
both regions pointing at one shared, replication-capable Redis.

### What HA actually buys, measured

The queue **survives** the node loss. n8n does **not** ride the failover out. A forced
failover on a live cluster promoted the replica in ~20s with queued executions intact,
and every main, worker and webhook pod exited and restarted while it happened. The
original version of this PR claimed the opposite; the live test is what corrected it.
See PR 3 for why, and for the knob that addresses it.

### Upgrade safety

Verified live, not reasoned about. A deployment applied on `main` and then repointed at
this branch planned `0 to add, 0 to change, 0 to destroy`, with three `has moved to
...[0]` lines. The Helm release did not re-render, which confirms the KEDA trigger
address rewrite is byte-identical on the default path.

Switching topologies **does** replace Redis: they are different resource types, so no
`moved` block can bridge them. README carries the drain-first procedure.

### Two decisions worth reviewing

- **The replication group identifier is `<cluster_name>-redis-rg`,** deliberately
  neutral rather than `-redis-ha`. PR 2 puts transit encryption on the same resource,
  so a topology-specific name would misdescribe it for half its callers, and
  `replication_group_id` forces replacement, so correcting it after release would cost
  early adopters their queue a second time. It must not change once merged.
- **It also carries a distinct suffix from the cluster's `cluster_id` on purpose.**
  ElastiCache shares one identifier namespace between cache clusters and replication
  groups. With a shared name the apply that flips the toggle destroys the old cache and
  then fails to create the replacement, leaving no queue backend at all. Found on a live
  cluster, and pinned by a test.
- **`at_rest_encryption_enabled = true`** is set in this PR rather than later, for the
  same ForceNew reason. Free on the AWS-managed key. The cluster resource has no
  equivalent argument, so the default path is untouched. Checkov `CKV_AWS_191` (CMK)
  is deliberately left open, matching how the module already treats RDS.

### Verification

- 111 `terraform test` runs, and the full CI matrix at the pinned versions
  (Terraform 1.9.8, tflint 0.53.0, terraform-docs 0.22.0) across all seven targets
- Live: baseline apply, no-diff upgrade, HA topology (001/us-east-1a primary,
  002/us-east-1b replica, AutoFailover + MultiAZ enabled), forced failover with the
  queue readable afterwards, external-Redis hook, and a clean teardown


jrx (approved):
Reviewed as a Terraform Registry module change and verified live on cluster `issue44` (us-east-1).

**Review:** additive and default-preserving; `moved` blocks absorb the count retrofit; the `-redis-rg` identifier split and create-time `at_rest_encryption_enabled` are correct ForceNew calls; all new checks/validations use the 1.9-safe `guard ? body : true` shape. Review fixes landed on the branch: `nullable = false` on the new toggle/port inputs (20eaf26) and softened threshold wording pending #77 (c667b25). Branch rebased onto current main, linear history, full local loop green (fmt, validate, 248 tests, tflint, docs check).

**Live verification:**
- Plan against a deployment applied earlier: `No changes` (upgrade safety re-confirmed after rebase)
- `redis_high_availability_enabled = null` from a real caller plans cleanly (nullable fix, the case mocks cannot express)
- Check block warns and the decoy `redis_host` is genuinely ignored
- HA flip: replication group `available`, failover + Multi-AZ enabled, nodes split us-east-1a/1b, old cluster destroyed in the same apply
- Forced failover: promotion ~20s, primary repointed, queue intact
- Smoke test 32/32 including execution via queue and KEDA scale 1→4 under 100-job load

PR #76, opened 2026-08-05
feat: encrypt the Redis queue backend in transit with an AUTH token

### LINKED ISSUE #41: feat: add redis_transit_encryption_enabled for ElastiCache TLS + auth token

### LINKED ISSUE #66: KEDA worker autoscaler loses Redis auth when redis_transit_encryption_enabled is set

buddy-n8n:
Closes #41. Closes #66.

Adds `redis_transit_encryption_enabled` (default `false`), which puts TLS and a
generated AUTH token on the ElastiCache queue backend, and teaches KEDA's Redis
triggers to speak to it.

### Why this supersedes two branches

`fix/issue-66-keda-redis-tls-auth` already contained all of
`feat/issue-41-redis-transit-encryption` and replaced its `KNOWN GAP` comment with a
working fix, so **`feat/issue-41-redis-transit-encryption` should be closed, not
merged.** The two also cannot ship separately: #41 alone puts a broken feature in main,
where enabling TLS silently stalls worker autoscaling.

Both of those branches also defined their own `aws_elasticache_replication_group.n8n`,
which would have collided with PR 1's. This PR collapses all of it onto **one**
resource.

### One resource, two independent features

`auth_token` exists only on `aws_elasticache_replication_group`, and AWS requires
transit encryption before AUTH can be enabled at all. So this variable and
`redis_high_availability_enabled` both select that resource, for unrelated reasons, and
either one alone is enough. They stay independent on it: encryption leaves the cache
single-node, availability leaves the endpoint plaintext. All four combinations are
pinned by tests.

### The KEDA half

Both triggers gain `enableTLS` and `passwordFromEnv: QUEUE_BULL_REDIS_PASSWORD`. TLS is
the half that has to land: without it KEDA opens a plaintext connection to a TLS-only
endpoint and hangs on `connection to redis failed: i/o timeout` before authentication is
ever attempted. Nothing crashes, the HPA just reports `<unknown>` and workers freeze at
their current replica count.

The token reaches KEDA by env-var reference, resolved against the worker pod's first
container via the `secretKeyRef` the chart already sets, so it never appears in the
ScaledObject and no `TriggerAuthentication` resource is needed. This depends on KEDA
being allowed to read Secrets outside its namespace, which its chart permits by default.

### New guard

`redis_transit_encryption_enabled` with `create_elasticache = false` is now a hard
plan-time error. Neither source branch had this. The module cannot put a token on a
Redis it does not manage, and the combination would otherwise render `tls = true` plus a
generated password against the caller's plaintext endpoint: an apply that succeeds and
then fails at runtime.

### Reviewer note

`auth_token_update_strategy` is gated on the same variable as `auth_token`, not set
unconditionally. The provider rejects it outright when there is no token
(`"auth_token_update_strategy": "auth_token" must be specified`), which breaks the
HA-only plan. Neither `terraform validate` nor `terraform test` catches this, because
the test framework's mocked provider skips the real provider's argument validation. A
live plan is what found it, and there is now an assert standing in for it.

### Verification

- 136 `terraform test` runs, full CI matrix at pinned versions
- **HA + transit encryption together, live on a real cluster** (`us-east-1`, 2 nodes
  across `us-east-1a`/`1b`, `TransitEncryptionMode: required`, AUTH enabled). Plaintext
  `PING` hangs with and without a token; TLS without a token returns `NOAUTH
  Authentication required.`; TLS with the token returns `PONG`, including with full
  certificate verification. KEDA `ScaledObject READY=True` with a numeric HPA and
  triggers carrying `enableTLS: "true"` and `passwordFromEnv`.
- **Forced failover with TLS on**: primary promoted across AZs, no container terminated,
  every pod logged `Recovered Redis connection`, queue intact afterwards.
- **AUTH token rotation** (the P1 finding): all five pods across main, worker and
  webhook processor were replaced, every new pod's `checksum/redis-auth-token`
  annotation equals `sha256` of the new token, and every pod's
  `QUEUE_BULL_REDIS_PASSWORD` matches `terraform output -raw redis_auth_token`. Both the
  old and new token answer during the roll, so there is no lockout window.
- **The three-step TLS migration**, end to end on a live replication group with a client
  connection held open across every step. No step interrupted service. `preferred` took
  17m27s with 1198 consecutive replies and zero errors; `required` took 8m18s in one
  Terraform apply, closing plaintext 131s in; the final rotation took seconds. The
  second rotation is what actually closes unauthenticated access, confirmed both ways.
- **Zero-diff check**: with this tree, an existing HA + TLS deployment plans
  `No changes.` Both new inputs are invisible at their defaults.


jrx (approved):
Approved as part of the 75→76→77 stack, restacked onto the rebased #75 and verified live on cluster `issue44`.

**Restack notes:** two semantic conflict resolutions worth knowing: (1) the `redis_high_availability_enabled` description now merges this PR's transit-encryption-aware text with the softened threshold phrase from #75; (2) commit "roll n8n pods when the AUTH token rotates" carried a stale copy of the old single-key image-pin block that main's #71 superseded — dropped in favor of the repository+tag merge block. Full loop green (fmt, validate, 273 tests, tflint, docs check).

**Live verification (TLS+AUTH enabled at replication-group create time, together with HA):**
- `transit_encryption_enabled = true`, mode `required`, AUTH on, `ROTATE` strategy — all confirmed on the created group
- Pods wired via `QUEUE_BULL_REDIS_TLS=true` and `QUEUE_BULL_REDIS_PASSWORD` from `n8n-enterprise-redis-secret`
- KEDA triggers carry `enableTLS: true` + `passwordFromEnv`; ScaledObject `READY=True` and scaled workers 1→4 in 45s under a 100-job queue — the exact path that used to hang against a TLS-only endpoint
- End-to-end execution via the encrypted queue completed successfully (smoke test 32/32)

Not re-verified live: the three-apply plaintext→preferred→required migration on an existing group (already verified in this PR's own live run; this round created the group with TLS from scratch). Follow-up worth a small PR: `nullable = false` on this PR's new inputs, same AGENTS.md rule as #75.

PR #77, opened 2026-08-05
feat: let callers extend n8n's Redis reconnect budget for failovers

buddy-n8n:
Adds `n8n_redis_timeout_threshold`, wiring the chart's `redis.timeout` to
`QUEUE_BULL_REDIS_TIMEOUT_THRESHOLD`. Defaults to `null`, which emits no key at all, so
the chart's own `10000` applies and existing releases see no Helm diff.

### What this resolves

PR 1 recorded that n8n restarts every pod during an HA failover, and that raising the
threshold to 30s only moved the exit later. The working hypothesis was a **stale-address
retry** in ioredis, which would have meant no threshold ever worked and only an upstream
fix would help.

**That hypothesis was wrong.** ioredis's `StandaloneConnector` rebuilds connection
options and passes the hostname to `net.createConnection` on every attempt, so it
re-resolves. Confirmed by watching a client recover on a genuinely different IP with no
restart.

The real mechanism: n8n never sets ioredis's `connectTimeout`, so it stays at 10s, and a
connect to a demoted primary hangs for that full 10s before failing. `makeRetryStrategy`
accumulates wall clock, so each failed attempt spends ~11.1s of the budget. **The
threshold is quantized:**

| You set | Real budget | Reconnect attempts before exit |
|---|---|---|
| `10000` (default) | 11.1s | 1 |
| `30000` | 33.2s | 3 |
| `60000` | 66.4s | 6 |

Reproduced with n8n's exact client options and a verbatim copy of its retry strategy
(two Redis instances, `/etc/hosts` flip standing in for the endpoint repointing,
`192.0.2.1` for the demoted primary so connects hang rather than being refused):

| Endpoint stale for | Recovered at | `10000` | `30000` | `60000` |
|---|---|---|---|---|
| 15s | 33.4s | exits 21.2s | survives | survives |
| 25s | 44.5s | exits 21.4s | **exits 43.4s** | survives |

The second row is the live failure reproduced: a 30s threshold fires **1.1 seconds
before** the connection would have come back. That is the whole reason raising it to 30s
looked like it did nothing.

### Why the default is left alone

Raising the threshold is a trade, not a free win. It is also what makes a pod fail fast
against a genuinely dead Redis. Shipping a number as the default would impose that trade
on everyone, so the variable is opt-in and the README carries the measurements.

### Future work: an upstream n8n issue worth filing

**This module can only mitigate, not fix.** The lever it has is the threshold; the
underlying problem is that `connectTimeout` is hardcoded at ioredis's 10s default with
no env var to change it, which is what makes the budget coarse in the first place.
Lowering it to 2s pulled recovery from 8.4s after the endpoint repointed down to 1.4s in
the same reproduction.

A draft issue for `n8n-io/n8n` is written up but **not filed**, proposing in order:

1. Set a short `connectTimeout` on the queue client (one line in `getOptions()`)
2. Stop counting connect-timeout time against the threshold, or document that it is a
   wall-clock budget rather than a count of attempts
3. Expose `QUEUE_BULL_REDIS_CONNECT_TIMEOUT`
4. Raise the default threshold, which at 10s cannot survive any managed-Redis failover

Two related observations for whoever picks it up:

- `QUEUE_BULL_REDIS_RECONNECT_ON_FAILOVER` does not apply here. It keys on `READONLY`,
  and this failure is `connect ETIMEDOUT`, where the connection does not survive at all.
- `QUEUE_BULL_REDIS_DNS_LOOKUP_STRATEGY` does not apply either: the `dnsLookup` override
  is set in `getClusterOptions` only, so it never reaches the regular client used for a
  non-clustered endpoint. Worth documenting upstream regardless.

### Verified against a real ElastiCache failover

`redis_high_availability_enabled = true` plus `n8n_redis_timeout_threshold =
60000`, forced failover on a live cluster. **No container terminated.** Every
pod logged `Recovered Redis connection` rather than exiting. The only two
terminations in the namespace predated the failover by two minutes and were
ordinary startup churn. Roles swapped and the sentinel key written beforehand
was still readable afterwards.

n8n's own counter shows the predicted quantum, measured rather than inferred:

```
Lost Redis connection. Trying to reconnect in 1s... (18.1s/60s)
Lost Redis connection. Trying to reconnect in 1s... (29.1s/60s)
Lost Redis connection. Trying to reconnect in 1s... (40.1s/60s)
Recovered Redis connection
```

Gaps of 11.1s and 11.0s, matching the model. A 30s threshold would have exited
at the 40.1s sample, ~11 seconds before recovery.

**One correction to the local reproduction:** the real endpoint stayed stale for
**48 seconds**, roughly double the worst case modelled, measured by resolving
the primary endpoint once a second from inside the cluster. CoreDNS caching plus
the endpoint TTL stretches the window well past the promotion. 60000 cleared it
with ~20s of headroom, from a single observed failover, so the README presents
it as a good default rather than a guarantee.

### Verification summary

- 123 `terraform test` runs, full CI matrix at pinned versions
  (Terraform 1.9.8, tflint 0.53.0, terraform-docs 0.22.0, checkov) across all
  seven targets
- Mechanism established by reproduction, then confirmed live as above

Known gaps: the write-outage probe misfired (`redis-cli --connect-timeout` does
not exist in `redis:7-alpine`), so queue durability was confirmed by sentinel
key rather than by a write timeline; only one failover was observed; and HA
combined with `redis_transit_encryption_enabled` is covered by plan-time tests
only, not live.


jrx (approved):
Approved as part of the 75→76→77 stack, restacked onto the rebased #76 and verified live on cluster `issue44`.

The threshold investigation in this PR is what corrected #75's original "raising the threshold only delays the exit" conclusion, and the live run here confirms it end to end.

**Live verification (`n8n_redis_timeout_threshold = 60000` on the HA+TLS group):**
- `QUEUE_BULL_REDIS_TIMEOUT_THRESHOLD=60000` rendered into the ConfigMap; default path stays a byte-identical no-op (`No changes` plan against a live deployment)
- Forced ElastiCache failover (`test-failover`, node group 0001): primary flipped 001→002, **zero pod restarts, zero `process.exit`** — every main, worker, and webhook pod logged the counting reconnect loop (`(7.1s/60s)`) and then `Recovered Redis connection`
- healthz 200 and ScaledObject `READY=True` throughout; smoke test with load 32/32 afterwards

That is the stack's headline behavior delivered: #75 alone measured a full pod-fleet restart on failover; with this threshold the fleet rides it out with the queue intact.

PR #78, opened 2026-08-05
feat: expose alb_ssl_policy for the ALB HTTPS listener

### LINKED ISSUE #43: feat: expose alb_ssl_policy for the ALB HTTPS listener
## Problem

The n8n Ingress defines the HTTPS listener (`alb.ingress.kubernetes.io/listen-ports` with `HTTPS = 443`, `ssl-redirect = 443`) but does not set `alb.ingress.kubernetes.io/ssl-policy`. The ALB therefore falls back to its default negotiation policy, which is not pinned by the module and can drift as AWS changes defaults. There is no ADR covering the ALB listener TLS policy (the security-posture ADR addresses database and Redis transport only), so this is an uncovered gap.

Operators with a compliance baseline (TLS 1.2 minimum, or TLS 1.3-only) currently cannot set the negotiated policy without forking the module, and cannot assert what policy is actually in effect from the Terraform configuration.

## Proposal

Add an optional `alb_ssl_policy` variable (string) wired to `alb.ingress.kubernetes.io/ssl-policy`. Default to a current, well-supported modern policy (e.g. `ELBSecurityPolicy-TLS13-1-2-2021-06`) so the negotiated policy is explicit and pinned rather than implicit. Document that callers can set any AWS-published ELB security policy name.

If preserving the exact current (unpinned) behavior is preferred, the default can instead be `null` (annotation omitted) — but pinning a modern policy is the safer default and makes the posture declarative.

## Notes

- Additive; the only behavioral change is that the negotiated TLS policy becomes explicit instead of the ALB default.
- Follows the curated-surface contract (dedicated, validated, documented variable).
- Validation can check the value against a `ELBSecurityPolicy-` prefix.
- Needs a `variables.tf` entry, wiring in `n8n.tf` (Ingress annotation), a `.tftest.hcl` assert, a CHANGELOG bullet, and a `terraform-docs` refresh.

buddy-n8n:
## Summary

The module-managed Ingress never set `alb.ingress.kubernetes.io/ssl-policy`, so the ALB HTTPS listener negotiated whatever the AWS Load Balancer Controller's own default happened to be, unpinned by the module and invisible in the Terraform configuration. This adds `alb_ssl_policy` (default `ELBSecurityPolicy-TLS13-1-2-2021-06`), validated against the `ELBSecurityPolicy-` prefix, wired into the Ingress annotations the same way `ingress_scheme` is: mergeable/overridable via `ingress_annotations`, with a plan-time warning if that override happens silently, and folded into the existing `create_ingress = false` tuning check.

## Linked issue

Closes #43

## Stability impact

- [ ] Patch-eligible — additive only (new optional input, new output, new resource that doesn't disturb existing state, bug fix, docs).
- [x] **Minor-only** — new optional input `alb_ssl_policy`, but its default is not inert: the AWS Load Balancer Controller's own default when the annotation is absent is `ELBSecurityPolicy-2016-08` (permits TLS 1.0/1.1). The first apply after upgrading sets the annotation explicitly to `ELBSecurityPolicy-TLS13-1-2-2021-06`, raising the ALB's negotiated TLS floor. This is an in-place annotation update on the existing Ingress/ALB (no replacement — live-verified below), but it can reject handshakes from TLS 1.0/1.1-only clients. Documented as an upgrade note in `CHANGELOG.md`; callers can set `alb_ssl_policy = "ELBSecurityPolicy-2016-08"` to defer the change.

## Checklist

- [x] `terraform fmt -recursive` is clean.
- [x] `terraform validate` passes at the module root.
- [x] `terraform test -verbose` passes at the module root (105/105) — also re-verified on every example (`small`, `medium`, `large`, `cloudflare`, `godaddy`, `split-ingress`), none of which are touched by this change.
- [x] `tflint --format compact` is clean at the module root (and every example).
- [x] Ran `terraform-docs markdown table --output-file README.md --output-mode inject .` and committed the refreshed README.
- [x] Added 5 plan-time assertions in `tests/defaults.tftest.hcl`: default applies, validator rejects a non-`ELBSecurityPolicy-` value, the annotation-override warning fires, the `create_ingress = false` tuning warning fires, and the existing annotation-merge test was adjusted since `ssl-policy` is no longer an "unopinionated" example.
- [x] Updated `CHANGELOG.md` under `[Unreleased]` with an upgrade note.
- [x] Conventional Commits style on the commit subject (`feat: ...`).

## Verification

Beyond the checklist above, this was live-tested end-to-end against a real EKS cluster in `us-east-1` (account `087003410722`, cluster `iss43dev`) since the annotation flows through a real AWS Load Balancer Controller, not just Terraform state:

- **Baseline**: `terraform apply` with the module default. `kubectl get ingress` showed `alb.ingress.kubernetes.io/ssl-policy: ELBSecurityPolicy-TLS13-1-2-2021-06`, and `aws elbv2 describe-listeners` confirmed the real ALB's 443 listener `SslPolicy` matched exactly.
- **Override**: `terraform plan -var alb_ssl_policy=ELBSecurityPolicy-TLS13-1-3-2021-06` planned `0 to add, 2 to change, 0 to destroy` — confirming an in-place update. After apply, the same ALB ARN reported the new `SslPolicy` within ~20s (LBC reconciliation).
- **Enforcement, not just labeling**: against the TLS-1.3-only override, `openssl s_client -tls1_3` completed a clean handshake (`Verify return code: 0`), while `-tls1_2` failed mid-handshake — confirming the ALB actually enforces the negotiated policy.
- **Regression**: the repo's `tests/scripts/smoke-test.sh` passed 26/26 (0 failures) against the live cluster — pod health, HPA/KEDA, Redis, task runners, Ingress routing, HTTP health checks all green.
- `checkov` findings unchanged vs `main` (28 pre-existing, unrelated findings; `soft_fail: true` regardless).


buddy-n8n:
Both cubic findings are valid, fixed in 13a75b7:

1. **variables.tf:189** — the description's example dropped the `ELBSecurityPolicy-` prefix on the FS family (`FS-1-2-*` → `ELBSecurityPolicy-FS-1-2-*`), which would have failed the validator if followed literally. Confirmed against AWS's actual published policy names (e.g. `ELBSecurityPolicy-FS-1-2-2019-08`) before fixing.
2. **n8n.tf:603** — the `create_ingress = false` tuning check compared against a hardcoded copy of the default. Pulled it into `local.alb_ssl_policy_default` (locals.tf) so the check derives from one place instead of duplicating the literal; `variables.tf`'s own default has to stay a literal (Terraform variable defaults can't reference locals), documented with a comment on the local.

`terraform fmt`, `validate`, `test` (105/105), `tflint`, and `terraform-docs --output-check` all re-verified clean after the fix.

jrx (approved):
Reviewed against the repo's AGENTS.md quality bar and Terraform Registry best practices.

- Additive input `alb_ssl_policy` with description, type, default, and prefix validation; wired to the `alb.ingress.kubernetes.io/ssl-policy` annotation on the module-managed Ingress.
- Conflict-warning `check` block mirrors the existing `ingress_scheme` pattern and stays 1.9-safe; the `create_ingress = false` inert-tuning check was extended correctly.
- Test coverage is complete: default pin, override, validator rejection, annotation-conflict warning, and `create_ingress = false` warning.
- CHANGELOG entry includes a thorough upgrade note for the runtime behavior change (TLS floor rises from the LBC default `ELBSecurityPolicy-2016-08` to TLS 1.2+, with an explicit opt-out path).
- README regenerated and `ingress_annotations` guidance updated in lockstep.
- Branch rebased onto current `main`, review-fixup commit squashed, single linear commit with original authorship preserved. Full local loop green post-rebase: fmt, validate, `terraform test` (129/129), tflint, `terraform-docs --output-check`.

PR #80, opened 2026-08-05
chore: commit multi-platform provider lock files for caching

### LINKED ISSUE #79: chore: commit multi-platform provider lock files for local and CI caching
## Problem

Provider dependencies are not locked in the repo. `.gitignore` line 13 ignores
`*.terraform.lock.*`, so every `terraform init` in the module root and in each
example resolves and downloads providers again from the registry. That means:

- No local provider cache can be keyed on a stable lock file hash, so repeated
  local runs of the dev loop (six examples plus the module root) re-download the
  same providers.
- CI has nothing to cache against either, since `actions/cache` for the plugin
  directory normally keys on the hash of `.terraform.lock.hcl`.
- Contributors on different machines can silently resolve different provider
  patch versions inside the same `~> 6.0` style constraints.

## Proposal

Commit `.terraform.lock.hcl` for the **example root modules** and generate the
hashes for multiple platforms, so the same lock file works on Linux CI runners
and on Apple Silicon laptops:

```bash
terraform providers lock \
  -platform=linux_amd64 \
  -platform=linux_arm64 \
  -platform=darwin_arm64
```

Intel macOS (`darwin_amd64`) is intentionally out of scope.

Scope:

- [ ] Drop or narrow the `*.terraform.lock.*` ignore rule in `.gitignore`
- [ ] Generate and commit `.terraform.lock.hcl` at the module root, keeping
      `lockfile: false` in `.terraform-docs.yml`
- [ ] Generate and commit `.terraform.lock.hcl` in `examples/small`,
      `examples/medium`, `examples/large`, `examples/cloudflare`,
      `examples/godaddy`, `examples/split-ingress`
- [ ] Cache the provider plugin directory in
      `.github/workflows/terraform-tests.yml`, keyed on the lock file hashes
- [ ] Document how to refresh the lock files when a provider constraint changes
      (a single `terraform providers lock` invocation per example)

## The module root

The module root is locked too, in addition to the examples.

Caveat to respect while doing it: per HashiCorp's [dependency lock file
docs](https://developer.hashicorp.com/terraform/language/files/dependency-lock),
the lock file belongs to a root module. Registry consumers of
`terraform-aws-n8n` never read a lock file committed at this module's root, they
resolve against their own. The only benefit here is caching the module-root
steps of the dev loop and CI.

That means `.terraform-docs.yml` must keep `lockfile: false` (see the comment at
lines 33-37). Otherwise the generated README reference table flips from provider
*constraints* to the pinned versions in the new root lock file, which both
misleads consumers and breaks the `terraform-docs --output-check .` job.

## Acceptance criteria

- A fresh clone can run the full local loop with providers served from cache
  after the first run, on both `linux_amd64` and `darwin_arm64`.
- `terraform-docs --output-check .` still passes and the README reference table
  still shows constraints, not pinned versions.
- CI (`fmt`, `validate`, `test`, `tflint`, `checkov`, `docs`) stays green.

buddy-n8n:
## Summary
- Commit `.terraform.lock.hcl` at the module root and in each `examples/*` root module, each locked for `linux_amd64`, `linux_arm64`, and `darwin_arm64`
- Wire `actions/cache` into the `validate`, `test`, and `tflint` jobs of `terraform-tests.yml`, keyed on the hash of all committed lock files, so CI reuses downloaded provider plugins across runs
- Narrow `.gitignore`'s lock-file rule so these files are tracked instead of stripped; also ignore Serena MCP's local project index (`.serena/`) and a personal scratch note on an unrelated upstream n8n issue
- Document the refresh command in `AGENTS.md` (`terraform providers lock -platform=... `) so contributors know how to regenerate a lock file after changing a `required_providers` constraint
- Fixed a real provider-version drift caught during verification: `examples/small` had locked `aws 6.56.0` while the module root and every other example locked `6.57.1` under the same `~> 6.0` constraint (transient registry drift during initial lock generation, exactly the failure mode this issue exists to prevent). Re-locked with `-upgrade` to match.

`.terraform-docs.yml` keeps `lockfile: false` so the README reference table still renders provider constraints, not pinned versions.

Closes #79

## Test plan
- [x] `terraform fmt -check -recursive` clean at root and all six examples
- [x] `terraform-docs --output-check .` clean at root and all six examples
- [x] `terraform validate` clean at root and all six examples
- [x] `terraform test -verbose` passing at root (129) and all six examples (small 1, medium 1, large 14, cloudflare 4, godaddy 5, split-ingress 12)
- [x] `tflint --format compact` clean at root and all six examples
- [x] `checkov` run locally: no new findings vs. the pre-existing baseline (soft_fail, unchanged since no `.tf` files were touched)
- [x] `actionlint` clean on `terraform-tests.yml`

### DIFF
--- a/.github/workflows/terraform-tests.yml
+++ b/.github/workflows/terraform-tests.yml
@@ -15,6 +15,11 @@ env:
   TFLINT_VERSION: "v0.53.0"
   TF_IN_AUTOMATION: "true"
   TF_INPUT: "false"
+  # Shared provider plugin cache across the validate/test/tflint matrices.
+  # `terraform init` fills this from the registry on a cache miss and reuses
+  # it on a hit, keyed below on the .terraform.lock.hcl files committed at
+  # the module root and in each examples/* root module.
+  TF_PLUGIN_CACHE_DIR: "${{ github.workspace }}/.terraform-plugin-cache"
 
 jobs:
   fmt:
@@ -95,9 +100,16 @@ jobs:
         with:
           terraform_version: ${{ env.TF_VERSION }}
 
+      - name: cache terraform providers
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.TF_PLUGIN_CACHE_DIR }}
+          key: tf-providers-${{ hashFiles('**/.terraform.lock.hcl') }}
+          restore-keys: tf-providers-
+
       - name: terraform init (no backend, providers only)
         working-directory: ${{ matrix.target }}
-        run: terraform init -backend=false
+        run: mkdir -p "$TF_PLUGIN_CACHE_DIR" && terraform init -backend=false
 
       - name: terraform validate
         working-directory: ${{ matrix.target }}
@@ -127,9 +139,16 @@ jobs:
         with:
           terraform_version: ${{ env.TF_VERSION }}
 
+      - name: cache terraform providers
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.TF_PLUGIN_CACHE_DIR }}
+          key: tf-providers-${{ hashFiles('**/.terraform.lock.hcl') }}
+          restore-keys: tf-providers-
+
       - name: terraform init
         working-directory: ${{ matrix.target }}
-        run: terraform init -backend=false
+        run: mkdir -p "$TF_PLUGIN_CACHE_DIR" && terraform init -backend=false
 
       - name: terraform test
         working-directory: ${{ matrix.target }}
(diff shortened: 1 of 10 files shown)


jrx (approved):
Reviewed against the Terraform Registry / Partner Premier quality bar and repo AGENTS.md policy.

- Lock files in all 7 root modules verified: 3-platform h1 hashes (linux_amd64, linux_arm64, darwin_arm64), versions consistent across roots, constraints match versions.tf and each example.
- Local verification: fmt, init + validate in all 7 roots, terraform test at module root (129 passed), terraform-docs --output-check clean everywhere.
- CI cache wiring (hashFiles key, restore-keys fallback, mkdir before init) follows the standard pattern; cache poisoning is a non-issue since init verifies against the committed lock file hashes.
- No module interface or behavior change; a lock file inside a published module is never read by consumers, and AGENTS.md now documents this correctly.
- Branch is cleanly rebased on main with a single linear commit, and all CI checks pass.

Approving.

PR #82, opened 2026-08-05
chore: align module with HVD conventions (CODEOWNERS, Taskfile, docs)

buddy-n8n:
## Summary

Closes #81. Aligns `terraform-aws-n8n` with HashiCorp Validated Design (HVD) module conventions:

- Adds `.github/CODEOWNERS` and `Taskfile.yml` wrapping the local fmt/validate/test/lint/docs loop.
- Restructures README's Prerequisites section into HVD's subsection convention.
- Adds `docs/upgrading-n8n.md`, `docs/pod-identity.md` (this module uses EKS Pod Identity, not IRSA), and `docs/helm-chart-coverage.md` (pinned to chart version 1.10.0, the module's `n8n_chart_version` default).
- Regroups `variables.tf`: `cluster_name`/`tags`/`namespace` moved into a new leading `Common` banner ahead of `Foundation inputs`, the local analog of HVD's `friendly_name_prefix`/`common_tags` block. No variables were renamed, so this is non-breaking.
- Cleans up markdownlint findings across docs and `variables.tf` descriptions, and scopes `.markdownlint.json` to this repo's actual intentional style.
- Adds a `CONTRIBUTORS` file.
- Adds `scripts/check-variable-banners.sh` (`task banners`) so the `# ── Section ──` grouping in `variables.tf`/`outputs.tf` can't silently rot to "no sections" as more PRs land: it fails a block with no banner above it, or a banner comment that drifts from the established format. It can't judge whether a variable is filed under the *right* banner, only that the convention itself holds, so AGENTS.md now documents the current banner list and how to place a new variable. Local-only for now (`task ci`), not wired into CI yet. Auditing the banner map for that table turned up one pre-existing case it can't catch on its own: `n8n_task_runner_request_timeout` sitting under `ElastiCache Redis`, moved into `Task runners`; non-breaking, no README diff.

## Test plan

- [x] `task fmt validate test lint banners docs markdown` all pass locally (module root + every example)
- [x] `terraform-docs --output-check .` confirms README wasn't left stale
- [x] `scripts/check-variable-banners.sh` verified against deliberately broken fixtures (missing banner, malformed banner) to confirm it fails correctly, then against the real files to confirm it passes
- [x] Rebased onto `main`, re-verified `task ci` (all 7 test suites, fmt/validate/lint/banners/docs/markdown clean)
- [x] Diffed the rebuilt commit history's tree against the pre-cleanup tree to confirm the commit reorganization introduced no content changes

PR #83, opened 2026-08-05
fix(security): curate checkov findings and hard-fail static analysis

### LINKED ISSUE #27: Curate checkov findings and flip `soft_fail` to `false`
## Context

This issue tracks curating the checkov findings and flipping `soft_fail` to `false` so the static-analysis check actually gates CI.
See PR #21 → section **E2** in the release plan.

## Current state

`.github/workflows/terraform-tests.yml` runs checkov in `soft_fail: true` mode:

```yaml
- name: checkov
  uses: bridgecrewio/checkov-action@v12
  with:
    soft_fail: true
    # Module surfaces low-severity findings (e.g. RDS deletion protection,
    # encryption knobs) that are intentional defaults for a getting-started
    # template. Flip soft_fail to false once we curate suppressions.
```

The source tree has exactly **one** `checkov:skip` annotation today (`CKV_AWS_293` on `database.tf` line 203 — RDS deletion protection, intentionally left off for example teardown). Every other finding checkov reports is silently absorbed.

## Why this needs to flip

`soft_fail: true` satisfies the "static analysis runs and reports the result" bar for Terraform Registry Partner Premier criteria, but it doesn't actually gate the module. A community contributor adding a new resource won't see checkov push back on missing encryption or over-permissive IAM. That's exactly the surface area where module quality erodes silently.

The honest first-release framing was "checkov runs, no findings block CI." For v0.2.0 we want "checkov runs, hard-fail on findings, every skip has a written justification."

## What needs to happen

1. Run `checkov -d .` locally against the module root with the v0.1.0 codebase. Capture all findings into a triage list.
2. Categorize each finding:
   - **Fix** — change the resource (e.g. add a missing encryption attribute, tighten an IAM policy).
   - **Skip with justification** — annotate the resource with `# checkov:skip=CKV_AWS_XXX:<one-line reason>`. The reason must explain *why* the default is intentional (e.g. "exposed via `var.db_storage_encrypted`; flipping default would break existing deployments without a migration").
   - **Suppress with policy** — for findings that don't make sense for this module class, suppress globally via `.checkov.yaml` or `--skip-check`, with a comment in the YAML explaining why.
3. Expected categories of findings based on a quick mental pass:
   - **IAM least-privilege** on `iam.tf` (cluster role, node role, LBC role, autoscaler role, S3 role). Many will be skipped-with-justification because they follow upstream AWS recommendations.
   - **Encryption knobs** on `database.tf`, `s3.tf`. Mostly already covered by `var.db_storage_encrypted` and S3 default encryption; document each.
   - **EKS public endpoint** access (`eks.tf`). The module's example posture is public-endpoint with restricted CIDR; document the rationale.
   - **RDS** — `CKV_AWS_293` already skipped (deletion protection). Probably more here.
4. Flip the workflow to `soft_fail: false`. Confirm CI passes.
5. Add `tests/scripts/check-checkov.sh` (optional) for contributors to run locally before pushing.
6. Update `CHANGELOG.md` and `README.md` to reflect the new static-analysis posture.

## Estimated effort

Half a day to a full day of focused work plus a careful review of every skip rationale. The skips are the part that needs scrutiny — they're the policy decisions the next maintainer (or auditor) will read.

## Stability impact

Workflow change only — no Terraform surface change, no behavior change for module consumers. Could technically ship in a v0.1.x patch, but lands here in v0.2.0 because (a) some of the *fixes* (not skips) may change resource defaults that move infra, which is minor-only per the stability contract, and (b) bundling with the other v0.2.0 work simplifies the release.

## Related

- v0.1.0 release plan: PR #21
- Existing skip: `database.tf:203` (`CKV_AWS_293`)
- Workflow: `.github/workflows/terraform-tests.yml`

buddy-n8n:
## Summary

Curates every checkov finding against this module for issue #27: each is either a real fix or an inline `checkov:skip=<ID>:<reason>` annotation on the resource it applies to. Flips `soft_fail: false` for the checkov job in `.github/workflows/terraform-tests.yml` so new findings gate CI going forward. No check is suppressed repo-wide: `.checkov.yaml` only sets a `skip-path` for the `tests/` directories.

Real fixes include EKS control-plane logging (all five log types, explicit retained log group), EKS secrets envelope encryption via a new module-managed KMS key (`var.eks_secrets_encryption_enabled`), configurable EKS API endpoint access with plan-time validation, Cluster Autoscaler IAM write actions scoped to this cluster's node group ASGs, daily ElastiCache snapshots, security group egress descriptions, and pod hardening on the `examples/large` pgbouncer deployment. Full detail is in the commit messages and `CHANGELOG.md`.

## What's new since this PR was first opened

This branch was originally rebased onto the then-open #82 (`chore/issue-81-hvd-alignment`) to pick up its Taskfile/CODEOWNERS/README restructuring and variable/output banner convention early. #82 has since merged, and this branch is now rebased onto post-merge `main` with a clean, single commit containing only this PR's own changes (see Status below).

That rebase (and the six other in-flight PRs that merged along the way: #65, #71, #75, #76, #77, #80) pulled in resources the original hard-fail gate had never seen, and surfaced a few new findings:

- **Aurora query logging split correctly.** `examples/large` had `log_statement`/`log_min_duration_statement` on the DB *cluster* parameter group, which Aurora PostgreSQL rejects (neither parameter exists in that family, confirmed against the RDS API, not just reasoned about). Split into the cluster group (`rds.force_ssl`) and a new DB parameter group attached to the writer/reader (the two logging parameters). Both groups' contents are now asserted at plan time, and their live values were confirmed against a real Aurora cluster (see Status below).
- **`CKV2_AWS_5` skip on `aws_security_group.redis`.** Same root cause as the existing `CKV2_AWS_30` skip on RDS: checkov doesn't build a graph edge between two `count`-expanded resources, so it can't see the security group's attachment. Verified by removing the `count` in isolation and watching the finding disappear.
- **`CKV2_AWS_50` skip on the Redis replication group's TLS-only shape**, where staying at one node without automatic failover is the intended design (encryption must not silently double the Redis bill).
- **`CKV_AWS_191` is now a real fix, not a skip:** see below.
- Two `variables.tf` banner sections added per #82's convention (`EKS cluster`, `S3`), a `checkov` task in `Taskfile.yml`, and the root RDS parameter-group test now asserts the family is derived from `db_engine_version` rather than hardcoded.

## New: opt-in customer-managed KMS key for Redis (closes `CKV_AWS_191` for real)

The original skip on `CKV_AWS_191` said a CMK for Redis was "worth having... tracked as a follow-up" rather than smuggled into a curation change. That follow-up is in this PR: `redis_kms_encryption_enabled` (default `false`) encrypts the ElastiCache replication group at rest with a module-created CMK (`aws_kms_key.redis`), the same pattern as `aws_kms_key.db`.

Defaults to `false`, not `true` like `db_storage_encrypted`: `at_rest_encryption_enabled` is unconditional already, so every deployment is encrypted at rest with an AWS-managed key regardless of this variable, and a CMK on top of that is an upgrade over an already-secure baseline, not a gap the default needs to close. `kms_key_id` only exists on `aws_elasticache_replication_group`, so this is a third variable, alongside `redis_high_availability_enabled` and `redis_transit_encryption_enabled`, that independently selects the replication group over the default single-node cluster: enabling it moves a deployment that asked for neither HA nor TLS onto the replication group at the same one-node cost, just a different resource type. `kms_key_id` is `ForceNew`, so flipping this on an *existing* deployment replaces Redis and drops the queue, same as the other two Redis topology variables.

`CKV_AWS_191` now passes checkov's static analysis unconditionally (verified locally): the check looks for `kms_key_id` being wired at all, not its runtime value, so the skip annotation was removed rather than updated. The ForceNew swap and the resulting CMK encryption were both confirmed against a live replication group (see Status below).

## New: fixed a real `rds.force_ssl` plan-drift bug, found during live validation

Live validation (below) surfaced a pre-existing bug unrelated to the checkov curation itself: `aws_db_parameter_group.n8n` (`database.tf`) and `aws_rds_cluster_parameter_group.aurora` (`examples/large/aurora.tf`) both set `rds.force_ssl` with no explicit `apply_method`. AWS treats this as a static parameter that only takes effect on `pending-reboot`, but Terraform's own default is `immediate`, so every `terraform plan` against an already-applied deployment showed a spurious in-place update on this resource forever. Both now set `apply_method = "pending-reboot"` explicitly. Confirmed against both live clusters: `terraform plan -refresh=true` now returns `No changes` where it previously showed permanent drift.

Also bumped `tests/scripts/smoke-test.sh`'s execution-completion poll from 15 x 2s (30s) to 30 x 2s (60s): a freshly booted worker's very first job pays a one-time cold-start cost connecting to Redis/the broker, which the old budget was too tight for and produced a false "still pending" warning on an execution that went on to succeed seconds later.

## Status: live-validated against two real clusters, rebased onto post-merge main

Local gate: `checkov` (0 failed / 197 passed / 69 skipped), `terraform fmt`, `terraform validate` (all targets), `terraform test` (294 root + 27 `examples/large` + 27 `examples/split-ingress`, all against mocked providers), `tflint`, `terraform-docs --output-check`, the variable/output banner check, and `markdownlint`.

Live validation, two clusters in a non-production dev account (both applied, checked, and torn down):

- **Root module at defaults:** EKS control-plane logging (all 5 log types) and secrets envelope encryption applied cleanly, both wired to the new module-managed KMS key. Cluster reachable under default endpoint access. Cluster Autoscaler scales node groups under the new IAM `ResourceTag` condition (forced a real scale-up, confirmed ASG desired capacity moved). ElastiCache snapshot retention applied. `redis_kms_encryption_enabled = false` (the default) produces zero plan diff; flipping it to `true` on the same live cluster produced exactly the documented ForceNew swap (single-node cluster to replication group) and encrypted the new replication group with the new CMK. Ran the repo's `smoke-test.sh` end to end (pod health, queue mode, KEDA, HTTPS, API, and a full webhook to worker execution) plus a manual webhook to HTTP-request-node workflow to confirm binary data actually lands in the module's S3 bucket.
- **`examples/large` (downsized for cost, same config shape):** Aurora parameter-group split applies cleanly, cluster and instance parameter groups both `in-sync` with the expected values (`rds.force_ssl` on the cluster group, the two logging parameters on the instance group). pgbouncer pods start healthy under the new `securityContext` and `resources`, and n8n connects through pgbouncer to Aurora with no errors.

Rebase onto post-merge `main`: done. Replaying this branch's full unsquashed history onto `main` re-triggered every commit #82 had already squash-merged, so the rebase was done as a single-commit replay of this PR's own diff instead (base stays `main`, no history from #82 remains). The rebase also surfaced one more real gap: this PR's two new `variables.tf` banners (`EKS cluster`, `S3`) hadn't been added to `scripts/check-variable-banners.sh`'s expected-banner list, so `task banners` failed post-rebase; fixed and folded in. Full local gate re-verified green after both the rebase and that fix.

## Test plan

- [x] Rebase onto `main` now that #82 has merged; confirm the diff shrinks to this PR's own commits, and re-verify the full local gate
- [x] Apply against a non-production cluster; confirm EKS control-plane logging and secrets encryption apply cleanly
- [x] Confirm the cluster remains reachable under the default endpoint access settings
- [x] Confirm Cluster Autoscaler still scales node groups under the new IAM `ResourceTag` condition
- [x] Confirm ElastiCache snapshot behavior
- [x] Confirm the `examples/large` pgbouncer pod still starts under the new `securityContext`
- [x] Confirm `redis_kms_encryption_enabled = true` applies cleanly against a live replication group and encrypts with the new CMK; confirm the default (`false`) produces zero plan diff against an existing deployment
- [x] Confirm the Aurora parameter-group split (cluster vs. instance parameters) applies cleanly in `examples/large`

Resolves #27.


jrx (approved):
Reviewed the Terraform changes and remediation. CI is green, all reported findings are addressed, and live validation against pr83.n8n-solutions.team passed.

PR #86, opened 2026-08-06
feat: land Redis transit encryption and configurable reconnect budget

buddy-n8n:
## Summary

Recovers the work from #76 and #77. Both show as **merged** on GitHub, but into a stacked PR's intermediate branch rather than `main`: #75's branch (`feat/issue-44-redis-ha`) was squash-merged into `main` first (`64f8154`), then #76 merged into that same branch (now orphaned relative to `main`), then #77 merged into #76's branch. Each PR's own "Merged" status was accurate for its stated base branch; none of those base branches was `main`, so the code never actually reached it.

This PR is that same diff, unmodified, applied cleanly onto `main`'s current tip (the exact commit #75's stack originally branched from) via a verified 3-way apply. No design changes, no rework: see the individual commit message for the full rationale carried over from #76/#77.

### `redis_transit_encryption_enabled` (default `false`, from #76)

Puts TLS and a generated AUTH token on the ElastiCache Redis backing the Bull queue, and wires KEDA's Redis triggers to use them (`enableTLS`, `passwordFromEnv`). Without the KEDA half, turning on encryption alone would silently stall worker autoscaling: KEDA opens a plaintext connection to a TLS-only endpoint and hangs before authenticating, and the HPA just reports `<unknown>` instead of erroring. Hard-errors at plan time when combined with `create_elasticache = false`, since the module can't put a token on a Redis it doesn't manage.

### `n8n_redis_timeout_threshold` (default `null`, no behavior change, from #77)

Wires the chart's `redis.timeout` to `QUEUE_BULL_REDIS_TIMEOUT_THRESHOLD`. Came out of debugging why n8n pods restart on every Redis failover: n8n never sets ioredis's `connectTimeout`, so a connect to a demoted primary hangs for the full default 10s before failing, and the retry budget is quantized in ~11.1s steps as a result. Verified live against two real Redis instances with a hosts-file failover simulation.

## Test plan

- [x] `git apply --3way` of the #76/#77 diff onto current `main`'s tip: all 9 files applied cleanly, zero conflicts
- [x] `terraform fmt -check -recursive`, `terraform validate`, `terraform-docs --output-check .`: all clean
- [x] `terraform test -verbose`: 280 passed, 0 failed (the `Check block assertion failed` warnings in the output are pre-existing and unrelated; the same count appears on unmodified `main`)


jrx (approved):
Live verification passed for Redis TLS/AUTH, credential rotation, KEDA integration, and HA failover with the 60-second reconnect budget. The measured staged TLS-to-HA transition is documented, and all Terraform CI checks pass.

Issue #88, opened 2026-08-07
[feature] Support customer-managed infrastructure: existing EKS, RDS, S3, and Redis


buddy-n8n:
### What problem are you trying to solve?

Right now, the module only knows how to create every piece of infrastructure it needs: the EKS cluster, RDS instance, ElastiCache Redis, and S3 bucket. Organizations that already run their own EKS clusters, centrally-managed databases, or a shared Redis fleet, or that operate under an SCP requiring an IAM permissions boundary on every role, cannot adopt the module at all: it would either fight their existing infrastructure or fail the first `CreateRole` outright.

The goal is to let n8n deploy onto infrastructure a platform or security team already owns, attaching to it rather than replacing it, while an unmodified deployment keeps working exactly as it does today.

### Proposed solution

`create_eks`, `create_database`, `create_elasticache`, and `create_s3_bucket` would each default to `true` (unchanged behavior) and could be set to `false` to point the module at an existing resource instead:

- **Existing EKS cluster** (`existing_eks_cluster_name`, gated behind an explicit `existing_eks_cluster_prerequisites_confirmed` attestation, since node capacity, Cluster Autoscaler auto-discovery tags, API reachability, and naming collisions on a shared cluster aren't checkable at plan time). `create_ebs_csi` would let a caller opt the EBS CSI addon/StorageClass out when the existing cluster already has one.
- **Existing RDS instance** (`db_host`), **existing ElastiCache/Redis** (`redis_host`, with TLS, AUTH, and Redis 6+ ACL usernames supported on the external path), and **existing S3 bucket** (`existing_s3_bucket_name`), the last one backed by a `data "aws_s3_bucket"` lookup that warns at plan time if the supplied bucket has no default encryption or no public-access block, so a caller finds out before deploying rather than during a security review.
- `install_lbc`, `install_cluster_autoscaler`, `install_metrics_server`, and `install_keda` would let a caller opt out of each Helm-installed controller in favor of one a platform team already manages via GitOps, while the underlying IAM role/policy/Pod Identity association stay unconditional since an externally-installed controller still needs them. A reference example should demonstrate this directly: a BYO cluster already running its own LBC and metrics-server, with the module skipping both. The minimum controller versions and CRDs the n8n workload expects (the KEDA `ScaledObject` API version, the LBC ingress class) should be documented in one place rather than left implicit, so an operator can check compatibility before pointing the module at existing controllers.
[...]

PR #89, opened 2026-08-07
feat: add existing-infra support, Redis HA/TLS, IAM boundary controls, and a controllers submodule

buddy-n8n:
## Summary

Lets the module attach to infrastructure a caller already owns instead of always creating it (existing EKS, RDS, S3, ElastiCache Redis, KMS keys, and every cluster controller), and hardens Redis for production use (HA, transit encryption, AUTH). Consolidates the pattern into one discoverable thing: a README "Customer-managed infrastructure" state-matrix section, `docs/customer-managed-infrastructure.md` for contributors, and four new runnable examples (`customer-managed-redis`, `customer-managed-s3`, `customer-managed-cluster`, `customer-managed-everything`).

Also extracts the AWS Load Balancer Controller, Cluster Autoscaler, metrics-server, KEDA, and the EBS CSI driver into a new `modules/controllers` submodule (state-safe via `moved` blocks), so an advanced caller on `create_eks = false` can invoke it directly with only the controllers they need.

Full change list and rationale are in the commit messages and `CHANGELOG.md`.

Tracked in #88.

## What is in here beyond the customer-managed toggles

Several things landed on this branch that the title does not obviously cover, listed so reviewers know to look for them:

- **`modules/controllers` submodule.** New public contract, invocable directly. Carries its own `nullable = false` and validations mirroring the root's, plus a documented install/destroy ordering contract for callers who use it without the root module.
- **`redis_key_prefix`.** Single source of truth for the Redis namespace: sets n8n's own prefix, Bull's `QUEUE_BULL_PREFIX` and the KEDA trigger's `listName` together. Required when two deployments share one external Redis, which the module cannot detect on its own.
- **Bring your own KMS key**, for RDS storage/Performance Insights/logs and for the S3 bucket, via `create_db_kms_key` / `create_s3_kms_key` / `db_logs_kms_key_enabled` and their ARN inputs.
- **External Secrets (Enterprise)** support, including a scoped Secrets Manager grant with an explicit `Deny` on anything the module manages itself.
- **IAM permissions boundary** wiring on every role the module creates.
- **`n8n_webhook_hpa_enabled`** and the `*_secret_ref` inputs, for callers who bring their own autoscaling policy or their own Kubernetes Secrets.
- **Terraform floor raised to `>= 1.11`** (breaking; see below).

## Breaking change: minimum Terraform version

`required_version` moves from `>= 1.9` to `>= 1.11` in all twelve declarations (root, `modules/controllers`, ten examples; eight of those in a `versions.tf`, `cloudflare` and `godaddy` inline in `providers.tf`).

Two reasons. `override_resource`'s `override_during` attribute, which `examples/customer-managed-redis` and `-s3`'s test suites need, arrived in 1.11 and is *silently ignored* before it, so a caller below the floor got a confusing assertion failure from `terraform test` rather than a version error. And CI had already moved to a single 1.15.8 pin, which left `>= 1.9` as a claim nothing exercised, on a repo that has already been bitten by exactly that kind of untested floor (`check` blocks did not short-circuit `&&`/`||` before Terraform 1.10, a hazard invisible on any newer local CLI).

**Upgrade:** bump the Terraform CLI to 1.11 or newer and re-run `terraform init`. No configuration edits, no state migration, no resource replacement, no input or output changed. On an older CLI, `init` fails immediately rather than planning anything, so there is no half-applied state to recover from.

## Review fixes (Cubic)

Two Cubic passes are resolved on this branch.

The first (37 findings against the initial push) is where the state-migration and IAM-collision fixes below came from. The ones worth a reviewer's attention:

- **Eighteen broken state-migration paths.** Every `count` this branch added to a previously-unconditional resource shifted its address to `[0]` with no `moved` block, so an upgrade would have planned a destroy and recreate of the live EKS cluster and node group, the S3 bucket holding binary execution data, and the Secrets the pods mount. Four *existing* `moved` blocks were also dead, sourcing the EBS CSI stack from a `[0]` index no released version ever wrote.
- **An IAM name collision that made `customer-managed-everything` unappliable.** It invokes `modules/controllers` directly *and* calls `module "n8n"`, whose own controllers call is always instantiated; both created identically named LBC and Cluster Autoscaler roles. Gating role, policy and attachment on the same `create_eks || install_<x>` predicate as the association fixes the collision and the stranded-role problem together.
- **A Helm value that was a silent no-op.** `webhookConfig.failurePolicy` does not exist in aws-load-balancer-controller chart 3.5.0 (verified with `helm show values`); the key is `webhookConfig.ingressValdationFailurePolicy`, upstream's typo included. The webhook stayed on its chart default of `Fail` and could still block Ingress deletion during teardown.
- **Cluster Autoscaler held no usable write permission on `create_eks = false`.** Its IAM condition keyed on `cluster_name` while the chart auto-discovers ASGs by `eks_cluster_name`.
- **External Secrets could not read a CMK-encrypted secret**, because Secrets Manager decrypts as the calling principal and the role had no `kms:Decrypt`.
- **KMS ARN inputs are now paired with booleans.** Gating a `count` on `arn == null` meant `db_kms_key_arn = aws_kms_key.mine.arn` failed the plan outright. Reproduced with a standalone probe before changing anything.

Three items in that pass were found while verifying Cubic's list rather than reported by it: two extra missing `moved` blocks, `s3_kms_key_arn` having no "set but ignored" check, and the README section that three variable descriptions and two check error messages pointed at never having existed.

The second pass (16 findings against that fix commit) is smaller in kind but two are real correctness/security bugs, not just polish:

- **`create_db_kms_key = false` and `create_s3_kms_key = false` demanded a BYO-key ARN even on paths that use no CMK at all** (external database, SSE-S3, existing bucket): the validation didn't check the same conditions its own description said it was gated on. Now it does.
- **The Secrets Manager `kms:Decrypt` grant was scoped only by `kms:ViaService`**, which permits decrypting *any* secret's key via Secrets Manager, not just the allow-listed ones, if some other policy or a secret's own resource policy ever grants it `GetSecretValue` on a secret this module never named. Added a `kms:EncryptionContext:SecretARN` condition pinned to the resolved allow-list so KMS enforces the same boundary `GetSecretValue`'s `Resource` list does.
- Instance-type validation rejected valid hyphenated types (`m7i-flex.2xlarge`) in two examples and in the root module's own `node_instance_type`. `modules/controllers`' four chart-repository validations accepted plain `http://` and hostless URLs. The rest is test/doc hygiene: a missing `expect_failures` run, a test with an uncaught cascading validation failure, several inaccurate comments, a check name that no longer matched its condition, a self-contradictory README sentence, and an ambiguous error message duplicated across two files.

## Test plan

Full local CI replay using CI's exact pinned toolchain (Terraform 1.15.8, tflint v0.53.0, terraform-docs v0.24.0, checkov 3.3.9), covering all 11 targets:

- `terraform fmt -check -recursive`: clean
- `terraform-docs --output-check`: clean on all 11
- `terraform validate`: clean on all 11
- `terraform test`: **582 passed, 0 failed**, across all 11
- `tflint`: clean on all 11
- `checkov`: **378 passed, 0 failed**, 196 skipped, via `tests/scripts/check-checkov.sh`. Note this now hard-fails, per #83
- `scripts/check-variable-banners.sh`: clean
- `markdownlint`: no new findings

New test coverage in this round: the LBC/Cluster Autoscaler IAM gate in both directions, the three KMS toggles' validations and their effect on the bucket's encryption configuration, six `run` blocks on variables a stale comment claimed could not be tested (checked with a throwaway probe file rather than taken at face value; the variables test fine, the comment was wrong), a `customer_managed_node_max` validation run that was missing its sibling coverage, and a dot-rejection run for `customer-managed-s3`'s `cluster_name`.

### Live validation

Nine deploy runs against a real AWS account across this branch's development, the latest two specifically to close the "not yet applied live" gaps this PR carried going into review.

Applied clean:

- **Greenfield (`create_eks = true`), full stack, 80 resources.** Exercised the `modules/controllers` extraction end to end: all four `helm_release`s (LBC, Cluster Autoscaler, metrics-server, KEDA), every controller IAM role, policy and Pod Identity association, the EBS CSI addon and the gp3 StorageClass all reached state through the submodule.
- **Customer-managed RDS, S3, and Redis (with and without AUTH)** layers, each pointed at live infrastructure from the greenfield stack. n8n came up reachable on each and workflows were created through its REST API afterwards, so these were validated at the application layer rather than just at apply.
- **`examples/customer-managed-everything`, end to end, own VPC/EKS/RDS/Redis/S3 stand-ins plus a direct `modules/controllers` invocation** (the one thing no earlier run had exercised), and the exact shape the 409 fix targets. It did not apply clean on the first two tries; both failures were real bugs the live apply caught that no plan-time check could have, and both are fixed on this branch now:
  - The stand-in RDS instance created a database named `n8n`, but the module hardcodes connecting to `n8n_enterprise` regardless of `create_database` (`n8n.tf`). Every n8n pod crash-looped on "database \"n8n_enterprise\" does not exist" until the stand-in's `db_name` was corrected.
  - The VPC's subnet tags keyed `kubernetes.io/cluster` off `var.cluster_name`, but the stand-in EKS cluster is actually named `"${var.cluster_name}-cm"`. The LBC (installed by the directly-invoked `modules/controllers`) could never auto-discover a matching subnet, so the ALB never provisioned and the example's own Ingress timed out waiting for it. Fixed by deriving both the cluster's `name` and the VPC's subnet tags from one local instead of two independently-typed string literals.

  With both fixed, the full stack came up clean: LBC, Cluster Autoscaler, metrics-server, KEDA and EBS CSI all `Running`, every n8n pod healthy, the Ingress got its ALB, and `n8n-main`'s `/healthz` returned `200` from inside the cluster. Confirms the `create_eks || install_<x>` gate actually prevents the 409 in the scenario it was written for: a direct `modules/controllers` call alongside `module "n8n"`'s own always-instantiated one.
- **The moved-block upgrade rehearsal.** Applied a stack at the `0.2.0` tag (no `modules/controllers`, flat root-level LBC/Cluster Autoscaler/EBS CSI resources), then repointed the same state at this branch's HEAD and ran `terraform plan`: **0 destroy, 0 replace** across all 18 `moved` blocks: every one matched its real, previously-applied resource and moved it cleanly into the new addresses. The only diffs were genuinely new resources and in-place attribute changes from unrelated feature work landed on `main` since `0.2.0` (KMS keys, S3 default encryption, security-group rule changes, chart version bumps), none of it caused by the refactor. This is the check a mocked-provider `terraform test` cannot perform at all, and it was this PR's largest remaining unverified claim going into review.

Failed live, by design, and that is where two of this PR's fixes come from:

- **Customer-managed EKS and everything-at-once**, in an earlier round, both hit `409 ResourceInUseException` creating a duplicate Pod Identity association for `aws-load-balancer-controller` and `cluster-autoscaler`. That is the origin of the `create_eks || install_<x>` gate, now confirmed fixed above.
- The ioredis reconnect-budget behaviour behind the Redis timeout work was likewise measured on a live cluster.

Observed live and worth knowing, not a bug in this PR: EKS Pod Identity has a propagation race. A pod created before its association finishes propagating to the Pod Identity Agent never gets credentials injected (`no EC2 IMDS role found`), and restarting that pod's container does not fix it: the mutating admission that wires in the credentials only runs at pod *creation*. Deleting the pod so the Deployment recreates it (after propagation completes) does. Hit this on `cluster-autoscaler` during the live runs above; self-healed within its own `CrashLoopBackOff` retry window in practice, but worth knowing if a fresh controller install looks unhealthy immediately after apply.

## Base

Based on current `main` (`e026dc5`), which already contains #82 and #83. No rebase outstanding.


jrx (approved):
## Review: approve

Full review + live validation campaign against real AWS (us-east-1). Details below.

### Static review
- Registry/HashiCorp conventions, input/output contracts, security posture, upgrade safety: all satisfied. Mandatory findings from the review pass fixed in 66e7c78 (stale LBC IAM header, three-platform provider locks for the four new examples, retired-Terraform-1.9 comment claims, n8n_encryption_key output null case).
- CI-equivalent loop green locally at the pinned versions: fmt, validate, terraform test (461 root + 121 example runs), tflint, checkov 3.3.9 (0 failed), terraform-docs, banner script.

### Live validation (all five stacks applied, verified, destroyed clean)
- **Upgrade migration**: examples/small applied at main, re-planned at branch tip: 0 to destroy, 32 moved, 4 documented in-place changes, converged. Migrated state later destroyed clean (81 resources, single pass).
- **Baseline**: smoke test 32/32 incl. end-to-end queue execution and the full KEDA scaling lifecycle (1 -> 10 -> 1).
- **customer-managed-cluster**: VPC postcondition + Pod Identity probe work; 409 hazard confirmed via CLI (ResourceInUseException) and the create_eks || install_* gate proven in both directions.
- **customer-managed-redis**: TLS+AUTH chain verified end to end (secretKeyRef wiring, KEDA authenticated queue reads, multi-main leader election, token output echo).
- **customer-managed-s3**: binary round trips on SSE-S3 and SSE-KMS; the documented kms:GenerateDataKey AccessDenied trap reproduced verbatim and resolved by s3_kms_key_arn alone.
- **customer-managed-everything**: direct modules/controllers invocation, zero-collision plan split, ordering contract held in both apply and destroy directions (no KEDA CRD finalizer deadlock), zero orphans.

### Fixed on the branch during live validation
- 58a77cc: customer-managed-cluster tagged subnets for the wrong cluster name (LBC discovery rejected them; ingress apply timed out)
- bb52b3b: both stand-in clusters orphaned a Never-expire control-plane log group on destroy
- 8759001: troubleshooting entry for the create_eks=false data-source deferral plan failure (+ the -refresh=false bucket-replacement footgun)

Zero AWS leftovers from the campaign.

PR #122, opened 2026-09-10
docs(readme): correct the customer-managed Redis TLS claim

buddy-n8n:
## What

The README's Redis in-transit-encryption chapter claimed `create_elasticache = false` was "not compatible" with `redis_transit_encryption_enabled`, and that the combination was "rejected at plan time." Neither has ever been true.

## Evidence

- `locals.tf`: `redis_tls_active` reads `var.redis_transit_encryption_enabled` unconditionally on both the module-managed and customer-managed paths. Only `redis_managed_tls_active` is gated on `create_elasticache`.
- `redis.tf`, `check "redis_tuning_requires_module_managed_elasticache"`: deliberately omits `redis_transit_encryption_enabled` from the inputs it rejects on the customer-managed path; its own message names that variable as the control for TLS posture there.
- `examples/customer-managed-redis` sets `create_elasticache = false` with `redis_transit_encryption_enabled = true`, and its test asserts the TLS path is exactly what the example exercises.

## Change

Rewrites the section to describe what the input actually means on the customer-managed path: a declaration that the caller's endpoint is TLS-only, which wires n8n and both KEDA queue-depth triggers to speak TLS. The module neither terminates nor verifies TLS itself. Calls out `redis_transit_encryption_mode` separately, since that one genuinely does not apply on that path (it's a property of the replication group the module manages).

Docs only, no behavior change. `CHANGELOG.md` has a matching entry under `Fixed`.

## Verification

- `terraform-docs --output-check .`: `README.md is up to date`.
- `markdownlint README.md`: clean on the edited range (pre-existing errors elsewhere are all inside the generated `BEGIN_TF_DOCS` block, unrelated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -488,6 +488,25 @@ this project adheres to the stability contract in
   call for this attribute on its own, and nothing about the running cache
   changes.
 
+### Fixed
+
+- **Docs:** the README's Redis in-transit encryption chapter claimed
+  `create_elasticache = false` was "not compatible" with
+  `redis_transit_encryption_enabled` and that the combination was "rejected at
+  plan time". Neither is true, and it has not been true on any released
+  version. `local.redis_tls_active` reads the input directly on both paths, and
+  `check "redis_tuning_requires_module_managed_elasticache"` (`redis.tf`)
+  deliberately omits it from the inputs it rejects on the customer-managed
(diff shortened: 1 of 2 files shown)


jrx (approved):
Reviewed the implementation and documentation against the customer-managed Redis TLS wiring. All required checks pass.

PR #130, opened 2026-09-14
fix(observability): scan and harden the opt-in redis_exporter checkov never saw

### LINKED ISSUE #120: checkov does not check kubernetes_*_v1 resources for pod security
## Problem

Checkov's pod security checks (`CKV_K8S_*`: run as non-root, drop capabilities, read-only root filesystem, memory limits, probes, and so on) only match the unsuffixed resource types (`kubernetes_deployment`, `kubernetes_service`, ...). They do not match the `_v1` variants.

This module uses `_v1` names throughout (`kubernetes_ingress_v1`, `kubernetes_service_account_v1`, `kubernetes_horizontal_pod_autoscaler_v2`, and since #105 `kubernetes_deployment_v1` / `kubernetes_service_v1`). Measured on the CI-pinned checkov 3.3.9:

- `examples/large/pgbouncer.tf` (`kubernetes_deployment`): 27 checks run
- `observability.tf` (`kubernetes_deployment_v1`): 0 checks run

So a green checkov run says nothing about pod security for most Kubernetes resources in the module. AGENTS.md presents checkov as a hard gate, which is misleading here. The exporter's hardening in `observability.tf` is correct only because it was written by hand, and a comment there is currently the only record of the gap.

## Options

1. **Document the gap in AGENTS.md** (static analysis section) so nobody reads "checkov passed" as pod security coverage. Cheap, should happen now.
2. **Report upstream** so checkov registers the `_v1` names as aliases. Right long-term fix, outside our control.
3. **Assert pod security fields in `terraform test`** for each `kubernetes_*_v1` workload, the way #105 already does for the exporter. Covers the gap until 2 lands.
4. Rename resources to the unsuffixed types. Not recommended: the provider treats them as different types, so it forces a destroy/recreate, and choosing a name so a scanner sees it is the wrong reason.

Suggested: 1 now, 2 as follow-up, 3 for anything added in the meantime.

## Context

Found during review of #105. Not caused by that PR; it affects every `_v1` Kubernetes resource in the module.

buddy-n8n:
Closes #120, with a corrected diagnosis.

## Problem

The CI checkov job never evaluates `kubernetes_deployment_v1.redis_exporter`. #120 (and the first version of this PR) attributed that to checkov not registering the `_v1`/`_v2` Kubernetes resource types. That is wrong: the 3.3.9 and 3.3.17 sources list `kubernetes_deployment_v1` in `supported_resources` for every `CKV_K8S_*` Terraform check, and `checkov -f observability.tf` on 3.3.17 draws the full 27 checks.

What hides the resource is count-gating. checkov evaluates `count` from variable defaults, answers every check on a count-0 resource with `UNKNOWN`, and omits it from the report. `redis_exporter_enabled` is `false` in the module defaults and in every example, so the default scan never reaches it. With `redis_exporter_enabled = true` it draws the same 27 checks as the always-on `examples/large/pgbouncer.tf` Deployment, and **three were failing untriaged**: `CKV_K8S_11` (CPU limit), `CKV_K8S_15` (pull policy), `CKV_K8S_43` (digest).

## Changes

- **`variables.tf`**: `redis_exporter_image` default is now `oliver006/redis_exporter:v1.90.0@sha256:a129504e…` (the multi-arch index digest, verified against Docker Hub and the registry). The digest makes the reference immutable, so the default `IfNotPresent` pull policy is safe. Fixes `CKV_K8S_15` and `CKV_K8S_43` on merit, matching the digest-pin precedent in `pgbouncer.tf`. Deployments on the default with the exporter enabled roll the pod once.
- **`observability.tf`**: `CKV_K8S_11` annotated at the resource as a deliberate trade (CPU uncapped so a CFS-throttled exporter cannot report late during an incident). Header comment corrected.
- **`tests/scripts/check-checkov.sh`**: now two passes. Pass 2 uses the new **`tests/checkov/opt-in.tfvars`** to turn every opt-in toggle on, and fails if it did not reach each address in `OPT_IN_RESOURCES`. Verified locally to fail on a removed `read_only_root_filesystem` and on a renamed address. No workflow change needed; CI already runs this script.
- **`tests/defaults.tftest.hcl`**: new run `redis_exporter_pod_spec_is_hardened_by_hand` pins single-container, security context, dropped capabilities with none added, a memory limit, the digest pin, and both probes on `/metrics:9121`. This covers fields checkov has no Terraform check for at all (`run_as_non_root`, the pinned UID, the probe target).
- **`AGENTS.md`**: "Known gap" rewritten around the real cause, with the two-pass gate and the rule for adding opt-in resources. No upstream checkov report is warranted.
- **`CHANGELOG.md`**: digest pin under `Security`, coverage fix under `Fixed`. **`README.md`** regenerated by terraform-docs.

## Verification

Rebased on `main` at 51c46e1 (#129, checkov pin 3.3.17). Terraform 1.16.2 (matches `TF_VERSION`).

- `terraform fmt -check -recursive`, `terraform validate`: clean
- `terraform test`: 578 passed, 0 failed
- `tests/scripts/check-checkov.sh`: pass 1 383 passed / 0 failed; pass 2 411 passed / 0 failed (exporter: 26 passed, 1 skipped)
- `terraform-docs --output-check .`, `tflint`, `markdownlint` (README, CONTRIBUTING, AGENTS, docs/*), `shellcheck`: clean


jrx (approved):
Approved. Root cause re-verified by measurement on checkov 3.3.17 and against the 3.3.9/3.3.17 sources; Oracle second opinion incorporated (digest pin over pull policy, two-pass gate, tightened asserts). CI green on 1d68ef5 with both checkov passes confirmed in the job log (383/0, 411/0). Only runtime change is the digest-pinned default of the same v1.90.0 image, verified against Docker Hub and the registry; no live apply needed.

PR #133, opened 2026-09-15
feat: add chart-values-diff task, link docs/versioning.md externally

buddy-n8n:
## Summary

Small follow-up to #129 (EKS Terraform version currency): two loose ends from that effort that weren't part of the actual chart/provider/engine version bumps.

- `tests/scripts/chart-values-diff.sh` + `task chart-diff CANDIDATE=<v>`: when bumping `n8n_chart_version`, you need to diff the candidate chart's `values.yaml` against the currently pinned one to see what changed (renamed keys, new required values, defaults that moved) before deciding whether this module's own inputs need updating to match. That was previously two manually-typed `helm show values` commands and a manual diff; this wraps it into one (`helm show values` against both versions via the same `oci://ghcr.io/n8n-io/n8n-helm-chart/n8n` reference, then `diff -u`). Informational only — it never writes anything and always exits 0, the same contract `check-version-drift.sh` already uses.
- `docs/versioning.md` (the full inventory of every version this module pins — Terraform providers, the five Helm charts it installs, the two database engines, the CI toolchain — and the bump policy for each) was only linked from `CONTRIBUTING.md` and `AGENTS.md`, both aimed at people modifying this module rather than people consuming it. `README.md`'s `Compatibility` section and `docs/upgrading-n8n.md`'s opening paragraph now link to it too, so anyone deciding whether to bump a provider or chart version in their own configuration can find the policy without first reading the contributor docs.

## Files (5)

- `CHANGELOG.md` [MODIFIED] (+10)
- `README.md` [MODIFIED] (+1 -1)
- `Taskfile.yml` [MODIFIED] (+5)
- `docs/upgrading-n8n.md` [MODIFIED] (+1 -1)
- `tests/scripts/chart-values-diff.sh` [ADDED] (+76)

## Validation

- Ran `tests/scripts/chart-values-diff.sh 1.10.0` against the live `oci://ghcr.io/n8n-io/n8n-helm-chart/n8n` registry: produced a real diff (the chart's `1.10.0` → `1.11.0` `listName`/`prefix` comment changes). Also confirmed the same-version guard (`chart-values-diff.sh 1.11.0` reports "already the pinned version; nothing to diff").
- `markdownlint README.md CONTRIBUTING.md AGENTS.md docs/*.md` — clean.
- `shellcheck tests/scripts/chart-values-diff.sh` — clean.
- `terraform-docs --output-check .` — clean (no `.tf` files touched).
- `terraform fmt -check -recursive` — clean.
- No `terraform test`/`tflint` re-run: no `.tf` file in this diff.


jrx (approved):
Reviewed against AGENTS.md and the Terraform Registry bar; no module interface, state, or provider changes. Follow-up commit fixed an invalid-YAML Taskfile desc, hardened the script's exit contract, consolidated read_default into tests/scripts/lib/tf-defaults.sh (now sourced by all three scripts), and pointed the chart-bump docs at task chart-diff. Rebased on main, all checks green.

